PR #3, opened 2020-09-08
missing `content` directive 

sayaikegawa:
Hi @mmcky ,

This PR fixes the issue that "contents" did not show the subtopics like below.
Screen Shot 2020-09-08 at 10 25 50 pm

"Contents" should be able to show the subtopics like below: (Note: this example does not refer to the about python lecture )
Screen Shot 2020-09-08 at 10 27 26 pm

### DIFF
--- a/source/rst/about_py.md
+++ b/source/rst/about_py.md
@@ -12,7 +12,9 @@ single: python
 
 # About Python
 
-## Contents
+```{contents}
+:depth: 2
+```
 
 ```{epigraph}
 "Python has gotten sufficiently weapons grade that we don’t descend into R


mmcky:
thanks @sayaikegawa -- would you mind moving this to a local branch for preview building?

sayaikegawa:
Hi @mmcky, 
Thank you for your comment. That's so helpful because I wondered why I cannot see preview buildings.

Re this "contents" issue, it seems to occur in all lectures. 
I'm wondering whether I should fix this issue in all lectures. This means that I'm wondering whether you would like to "know" particular types of errors that we need to do by hand in this test case or you would like to "fix" all errors in this test case.

If your purpose is the latter, I would like to fix this issue in all lectures at once.

mmcky:
thanks @sayaikegawa for a recurring issue like this perhaps document:

1. what currently happens in the source files
2. what should happen in the source files

and lodge this as an item #4

mmcky:
@jstac if our theme is similar to `sphinx-book-theme` with local contents on the right hand margin (cc: @drdrij) -- we would want to remove these local `contents` listings right?

DrDrij:
@mmcky I haven't factored in a "local" toc in the prototype so far (I'll have a demo up by Friday).
I think for a user, if a local toc is required for jumping to a section it's useful at the top - the was we have it on lectures currently.
The right column is handy while scrolling but I think in the long run having it permanently visible does not add to user readability.

mmcky:
roger that - thanks @DrDrij

jstac:
I'm in favor of a minimal, uncluttered design.  That's even more important than usual when you're teaching maths...

mmcky:
Support for `contents` directives has now been added in https://github.com/mmcky/sphinxcontrib-tomyst/pull/42

mmcky:
Fixed in #8

PR #14, opened 2020-09-14
monospaced_text_err

sayaikegawa:
Hi @mmcky ,
This PR fixes monospaced text error.
This error occurs in other lecture too. Especially, in the list.

[Here](https://5f5f3879598fe1455097dc6b--epic-agnesi-957267.netlify.app/functions.html#adding-conditions), monospaced text has redundant ">".
Screen Shot 2020-09-14 at 10 46 07 pm

### DIFF
--- a/source/rst/functions.md
+++ b/source/rst/functions.md
@@ -250,8 +250,8 @@ Notes
 
 * We are passing the argument `U` as a string, which is why we write it as `'U'`.
 * Notice that equality is tested with the `==` syntax, not `=`.
-    * > For example, the statement `> a = 10`>  assigns the name `> a`>  to the value `> 10`> .
-    * > The expression `> a == 10`>  evaluates to either `> True`>  or `> False`> , depending on the value of `> a`> .
+    * > For example, the statement `> a = 10`  assigns the name `> a` to the value `> 10`.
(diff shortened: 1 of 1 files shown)


source/rst/functions.md
```diff
@@ -250,8 +250,8 @@ Notes
 
 * We are passing the argument `U` as a string, which is why we write it as `'U'`.
 * Notice that equality is tested with the `==` syntax, not `=`.
-    * > For example, the statement `> a = 10`>  assigns the name `> a`>  to the value `> 10`> .
-    * > The expression `> a == 10`>  evaluates to either `> True`>  or `> False`> , depending on the value of `> a`> .
+    * > For example, the statement `> a = 10`  assigns the name `> a` to the value `> 10`.
```
mmcky, line 253: the `>` should probably be removed which is causing a comment or note like admonition in markdown syntax

mmcky: The source `rst` is:

```rst
* Notice that equality is tested with the ``==`` syntax, not ``=``.

    * For example, the statement ``a = 10`` assigns the name ``a`` to the value ``10``.

    * The expression ``a == 10`` evaluates to either ``True`` or ``False``, depending on the value of ``a``.
```

mmcky:
This is now fixed from changes in the source RST in #19

PR #65, opened 2020-10-07
✨ NEW: Adding plugins config

AakashGfude:
@mmcky you think adding menubar like this will be a nice temporary solution?
using a `plugins_list` config variable and passing the js file of the plugin to it. Then loading the whole js in the template.


mmcky:
@AakashGfude that looks like a good short term solution. 

It would be nice to have the `menubar` in a repo at some point to have a single point of update. But maybe that is best done when we build more `theme` infrastructure?

mmcky:
assuming this config works of course `:-)` -- doesn't look like the menubar is in the `preview`.

We could just have the `<script>` in the `template` folder right? And then reference the assets bucket?

AakashGfude:
aah yeah. should have made this a draft PR. the functionality in `qe-theme` is still at works.

AakashGfude:
mostly made this PR beforehand. to get validation about this approach from you :)

AakashGfude:
@mmcky  the menubar should be in the preview now.

mmcky:
nice work @AakashGfude

PR #73, opened 2020-10-09
[matplotlib] update default figuresize for lecture

mmcky:
Update matplot lib default figsize


mmcky:
https://matplotlib.org/3.1.1/users/dflt_style_changes.html#figure-size-font-size-and-screen-dpi

mmcky:
Hmm -- this doesn't seem to be taking affect. Perhaps the theme is sizing the output images?

PR #75, opened 2020-10-09
[functions] Migrate exercise to use sphinx-exercise

shlff:
Hi @mmcky , this PR converts [exercises and solutions of lecture functions](https://quantecon.github.io/lecture-python-programming.myst/functions.html#exercises) using ``sphinx-exercise``'s syntax.

I noticed two weird things in the following preview:
- for Exercise part, the name for exercises are ``Exercise 4.1``, ``Exercise 4.2``,  ``Exercise 4.3``, instead of ``Exercise 1``, ``Exercise 2``, ``Exercise 3``.   
- for Solution part, it fails to execute all ``python cells``!

For the issue in exercise numbering, I tried to use ``nonumber`` option. It just shuts down the whole numbering for exercises, but I don't find a way to change the names of exercises.


mmcky:
We are doing a major refactor of `sphinx-exercise` so once that is done we can use this as a test case.

mmcky:
@jstac I think we could enable `sphinx-exercise` for the lectures more generally now that `latex` is supported. 

While `code-cell` isn't directly enabled what do you think about this approach for the interim:

https://617656cde2d4e2288538c01a--epic-agnesi-957267.netlify.app/functions.html#solutions

I think it looks OK with code-cells following the solution block (so that it executes).

jstac:
I think it's very sensible `:-)`

I wonder how we can get this done though.  We are short on RA money.  Any thoughts?

Do we have Sloan money we can spend?

mmcky:
@jstac there is a great case in `writing_good_code` that shows we really need the gated directives (ideally). It has a lot of `code-cell` directives mixed in with `text` and hard to make it look cohesive in the above approach. 

We could either:

- [x]  migrate the lectures that contain straight forward cases as this is the majority of cases
- [x] put this on hold until we have gated directives (~1-2 months)

mmcky:
Closing in favour of #176

PR #76, opened 2020-10-12
Add logos in panels on the front page

mmcky:
This uses `myst` panels to add sloan and quantecon logos


mmcky:
Closing this as won't be needed with the move to use just `quantecon` logo. 

Couldn't get panels to work though.

PR #77, opened 2020-10-13
[MAINT] update environment to include build packages and remove from CI scripts

mmcky:


mmcky:
@DrDrij this updates the environment to include all the needed build tools

PR #78, opened 2020-10-19
Remove menubar from plugins list

DrDrij:
This removes the QuantEcon menubar script from being added to the builds.

The `banner.png` didn't come across with the new source as it was on the TOC page, replaced by `intro.md`.

### DIFF
--- a/lectures/_config.yml
+++ b/lectures/_config.yml
@@ -12,4 +12,3 @@ sphinx:
   config:
     html_theme: quantecon_book_theme
     html_static_path: ['_static']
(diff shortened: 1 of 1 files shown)


lectures/_config.yml
```diff
@@ -13,4 +13,3 @@ sphinx:
     html_theme: quantecon_book_theme
     html_static_path: ['_static']
     html_theme_options:
```
mmcky: hey @DrDrij we need to remove `html_theme_options` if there is nothing else nested below it.

PR #100, opened 2020-11-26
ENH: Build download notebooks using sphinx-tojupyter

mmcky:
This PR adds `sphinx-tojupyter` and enables support for `download notebooks`. 

These notebooks are built from the `sphinx.ast` and are modified from source notebooks in that:

1. they contain only basic `markdown` markup which makes them more readable in `jupyter notebook` or `jupyterlab` (for the time being until `myst` is rendered via `javascript`)
2. links are updated for context to point back to the website for external links (i.e. documents)


mmcky:
**Note:** this pr has the infrastructure for `pdf` production via `sphinx-tojupyter` but this is not enabled yet due to needing debug.

PR #107, opened 2020-12-03
Adjust code block to syntax highlighted literals

mmcky:
This improves compatibility with `sphinx-tojupyter`


mmcky:
this isn't needed as literal blocks are still added as code cells using `sphinx-tojupyter`

PR #116, opened 2020-12-06
[numba]_fix_imports

shlff:
Good morning, @jstac and @mmcky ,

I noticed an issue in lecture [numba](https://python-programming.quantecon.org/numba.html#compiling-classes):
- the import ``from numba.experimental import jitclass`` is not working 
- it returns an error message ``ModuleNotFoundError: No module named 'numba.experimental'``

This PR fixes this issue by replacing
- the original imports ``from numba import float64`` and ``from numba.experimental import jitclass``
- with modified imports ``from numba import float64, jitclass``

Do you think it is a good idea?

### DIFF
--- a/lectures/numba.md
+++ b/lectures/numba.md
@@ -286,8 +286,7 @@ created in {doc}`this lecture <python_oop>`.
 To compile this class we use the `@jitclass` decorator:
 
 ```{code-cell} python3
-from numba import float64
-from numba.experimental import jitclass
+from numba import float64, jitclass
 ```
 
 Notice that we also imported something called `float64`.


jstac:
Good catch @shlff , nice work.  

@mmcky I remember this change was made a few months back.  Do you think it's an issue in some of our other lectures that use numba?

mmcky:
@shlff are you testing this locally? You may want to update your `anaconda` if so.

It seems to be fine on the live site: https://python-programming.quantecon.org/numba.html

It should be `numba.experimental` to support future releases of `numba`

shlff:
Thanks for your comments, @jstac and @mmcky .

As @mmcky suggested, I had this issue because the version of my ``numba`` is 0.48.0, while they relocate ``jitclass``  between versions 0.48 and 0.49. 

Now I upgrade my numba, so this issue has been solved. 

I will close this PR.

PR #122, opened 2021-01-15
[numpy, scipy] Edits and Corrections

Horace510:
@shlff @jstac 
I added one useful feature of the Numpy to the file, it was one of the suggestions I made in my email.
Thanks!


lectures/scipy.md
```diff
@@ -442,7 +443,7 @@ We leave you to investigate the [set of available routines](http://docs.scipy.or
 (sp_ex1)=
 ### Exercise 1
 
-Previously we discussed the concept of {ref}`recursive function calls <recursive_functions>`.
+In Chapter 16, we will discuss the concept of {ref}`recursive function calls <recursive_functions>`.
```
jstac: @Horace510 , we need to avoid using numbers like 16 here because they can easily become wrong if the document changes.  You need to use a reference.  See here: https://jupyterbook.org/reference/cheatsheet.html#reference-documents

lectures/scipy.md
```diff
@@ -63,7 +63,8 @@ from numpy.fft import fft, ifft
 from numpy.lib.scimath import *
 ```
 
-However, it's more common and better practice to use NumPy functionality explicitly
+However, it's more common and better practice to use NumPy functionality explicitly. One good feature of Numpy is it has a faster processing time than other Python libraries.
```
jstac: @Horace510 : The statement you added is not precise enough.  Sometimes other libraries are faster.  (Also, Numpy should be NumPy.)

Also, the point here is that you can bring in NumPy implicitly, via SciPy, or explicitly.  It's better to be explict.  

How about adding ", rather than implicitly via SciPy" ?

lectures/scipy.md
mmcky: @Horace510 are you able to update this?

mmcky: I have made an update to resolve this

jstac:
Thanks @mmcky , this is good to go.

jstac (approved):

mmcky:
thanks @jstac

PR #126, opened 2021-03-10
MAINT: Update to use anaconda=2020.11

mmcky:
- [ ] setup a mechanism to do a clean run (without `cache`) for PR's like this


mmcky:
I have checked this locally against this new environment.

PR #138, opened 2021-05-06
MAINT: update quantecon-book-theme=0.2.0

mmcky:
This PR updates to `quantecon-book-theme=0.2.0`

PR #139, opened 2021-05-06
[pdflatex] Enable book style build of PDF via LaTeX

mmcky:
This PR brings book style PDF support via LaTeX

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,17 @@ jobs:
           python-version: 3.8
           environment-file: environment.yml
           activate-environment: lecture-python-programming
+      - name: Install latex dependencies
+        run: |
+          sudo apt-get -qq update
+          sudo apt-get install -y     \
+            texlive-latex-recommended \
+            texlive-latex-extra       \
+            texlive-fonts-recommended \
+            texlive-fonts-extra       \
+            texlive-xetex             \
+            latexmk                   \
+            xindy
       - name: Display Conda Environment Versions
         shell: bash -l {0}
         run: conda list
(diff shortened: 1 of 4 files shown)


mmcky:
@AakashGfude this doesn't seem to be picking up that there is a `pdf` available. How is the `name` of the file determined? I suspect i also need to update `_config.yml`

mmcky:
This is presenting a settings button for the `live` build launcher. With one option it should just retain the single launch button. I think maybe my `_config` isn't setup correctly on this either?

Screen Shot 2021-05-06 at 10 57 02 am

AakashGfude:
> This is presenting a settings button for the `live` build launcher. With one option it should just retain the single launch button. I think maybe my `_config` isn't setup correctly on this either?
> 
> Screen Shot 2021-05-06 at 10 57 02 am

Right @mmcky, we need to put only one launcher in `_config`. at present it has all of them https://github.com/QuantEcon/lecture-python-programming.myst/blob/master/lectures/_config.yml#L31

AakashGfude:
> @AakashGfude this doesn't seem to be picking up that there is a `pdf` available. How is the `name` of the file determined? I suspect i also need to update `_config.yml`

It gets the name from the `latex_documents` key. 
The reason it is not detecting here is that I am checking for the presence of pdf before showing it. SInce latex is run after HTML in the ci. When building HTML, pdf is not present. 
What I did earlier in the test PR was to move the latex build before HTML. https://github.com/QuantEcon/lecture-python.myst/pull/127/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR31

mmcky:
Hmm one interesting issue is the `cache` that I have on `github actions`. We may need a mechanism to rebuild `clean`

mmcky:
thanks @AakashGfude this is looking nice. Could you take a look and let me know what you think. 

- [x] needs to be merged after #138

PR #143, opened 2021-05-18
[python_by_example] Remove custom matplotlib default figsize style in plt.rcParams

mmcky:


mmcky:
Decided not to implement this so will close.

PR #151, opened 2021-06-23
MAINT: update anaconda=2021.05

mmcky:


mmcky:
This build error is not reproducible locally which suggests it is a `github actions` environment issue. 

- [ ] test `anaconda=2021.05` on its own

PR #159, opened 2021-07-26
MAINT: Update front page notices

mmcky:


mmcky:
@jstac here is the proposed amendment. There is however a "powered by jupyter-book" in the right hand margin. We could remove this entirely?

jstac:
Ah, good point @mmcky :-)

In that case I think we can just drop "These lectures are built using Jupyter Book."   It's redundant.

Thanks again, otherwise all good.

PR #182, opened 2022-05-20
TEST: Testing with Thebe

mmcky:
This looks at enabling `thebe` on our lectures


mmcky:
- [x] enabled binder https://github.com/QuantEcon/lecture-python-programming.myst/pull/183

mmcky:
It appears the destination file needs to be `ipynb` rather than `md` file (same as the source file)which is returning a 404. Our notebooks are hosted in a different repository. It can connect to it but the file extension is hard coded in the theme it seems. So the notebook isn't launching and a `404` is returned. 

- [ ] may need to support `thebe` in the `quantecon_book_theme` to fit the publishing context (cc: @AakashGfude) with the use of `html_theme_options`
- [ ] report issue when using `launch_buttons` as `jb` config and then setting some `html_theme` options via the sphinx configuration. The `jb` configuration is dropped and replaced.

mmcky:
putting this on hold. I suspect there are some system issues with `mybinder`. I have tried a lot of different configurations but nothing is working reliably at the moment.

mmcky:
OK after a day of testing and debugging the `build` cache still seems to reset a LOT and not reliable with `mybinder`

mmcky:
Looking at the source code generated by `html` if you click on the binder button for `matplotlib` it requests

```
https://mybinder.org/v2/gh/QuantEcon/lecture-python-programming.myst/sphinx-thebe?urlpath=tree/lectures/matplotlib.md
```

which appears to want to `build` every time it is clicked for any new `session` on the website -- and this build process fails a lot on `mybinder` at the moment and is very slow. I am not sure why the cached (previously built versions) aren't getting picked up. 

It appears to bring up the cached version with

```
https://mybinder.org/v2/gh/QuantEcon/lecture-python-programming.myst/sphinx-thebe
```

it sometimes works and sometimes doesn't connect. If I use a private `safari` window then it doesn't work and fetches `thebe` from `cdn` network and wants to build a new image (according to the feedback after the live button is pushed)

mmcky:
Closing as this config is now stale.

PR #184, opened 2022-06-14
ENH: Adding Google Colab Support

AakashGfude:
This PR adds Google Colab as a live launch option. 

It pulls the notebooks from the same source as `mybinder` which is `lecture-python-programming.notebooks`


environment.yml
```diff
@@ -1,4 +1,4 @@
-name: quantecon
+name: lecture-python-programming
```
mmcky: let's revert this as we have a universal `quantecon` environment now.

AakashGfude: oops niceeeee. okay will do.

mmcky:
@jstac I have given this too much time already -- so I think I need to leave the `google collab` testing for now. However the list of packages is larger than I thought including items like plotly

https://github.com/QuantEcon/lecture-python-programming.myst/blob/0d57a65c66588917cbaa6e6b4e6f80628cfae505/google-colab-requirements.txt

I'll park this work in a separate branch in case we come back to it. But essentially the google cola requirements can't be satisfied from `pypi` anymore. I guess that's b/c it is largely based on packages from 2020.

mmcky:
thanks @AakashGfude

PR #220, opened 2022-09-18
FIX: Fix Hint for Exercise 3.2

### LINKED ISSUE #219: Fix hint in Ex 2
In Ex 2 of https://python-programming.quantecon.org/python_by_example.html, this hint is not used "For the legend, noted that the expression 'foo' + str(42) evaluates to 'foo42'."

We should replace it with hint related to formatted strings (e.g. `f'foo{var}`)

HumphreyYang:
Hi @jstac,

This PR fixed #219. Please kindly review this PR and merge it if it looks good to you.

Please let me know if there is anything I can improve.

Many thanks in advance.

### DIFF
--- a/lectures/python_by_example.md
+++ b/lectures/python_by_example.md
@@ -525,7 +525,7 @@ If you can, add a legend, to help distinguish between the three time series.
 :class: dropdown
 
 * If you call the `plot()` function multiple times before calling `show()`, all of the lines you produce will end up on the same figure.
-* For the legend, noted that if `var = 42`, the expression `f'foo{var}'` evaluates to `'foo42'`.
+* For the legend, noted that suppose `var = 42`, the expression `f'foo{var}'` evaluates to `'foo42'`.
 ```
 
 ```{exercise-end}
--- a/lectures/python_by_example.md
+++ b/lectures/python_by_example.md
@@ -524,7 +524,10 @@ If you can, add a legend, to help distinguish between the three time series.
 Hints:
 
 * If you call the `plot()` function multiple times before calling `show()`, all of the lines you produce will end up on the same figure.
-* For the legend, noted that the expression `'foo' + str(42)` evaluates to `'foo42'`.
+* For the legend, noted that if `var = 42`, the expression `f'foo{var}'` evaluates to `'foo42'`.
+
+    Note: `f'foo{var}'` is an example of [f-String](https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings), which allows you to use `{}` to contain an expression. The contained expression will be evaluated, and the result will be placed into the string.
+
 ```
 
 ```{exercise-end}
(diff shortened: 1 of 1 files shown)


jstac:
Nice work @HumphreyYang , many thanks.

Perhaps the note should use https://jupyterbook.org/en/stable/content/content-blocks.html#notes-warnings-and-other-admonitions ?

lectures/python_by_example.md
```diff
@@ -524,7 +524,10 @@ If you can, add a legend, to help distinguish between the three time series.
 Hints:
 
 * If you call the `plot()` function multiple times before calling `show()`, all of the lines you produce will end up on the same figure.
-* For the legend, noted that the expression `'foo' + str(42)` evaluates to `'foo42'`.
+* For the legend, noted that if `var = 42`, the expression `f'foo{var}'` evaluates to `'foo42'`.
+
+    Note: `f'foo{var}'` is an example of [f-String](https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings), which allows you to use `{}` to contain an expression. The contained expression will be evaluated, and the result will be placed into the string.
```
mmcky: @AakashGfude can we use `note` admonitions with `quantecon-book-theme`?

HumphreyYang: Hi @mmcky, I have added note and hint admonitions. Could you have a look at the latest deployment and see if it looks good to you as well?

Many thanks.

mmcky:
hey @AakashGfude it looks like the styling is inherited

Screen Shot 2022-09-21 at 7 44 51 pm

lectures/python_by_example.md
mmcky: Thanks @HumphreyYang I think they look pretty good. Just not sure we need to dropdown class on the hint. Let's leave it as it is and run this by @jstac.

lectures/python_by_example.md
```diff
@@ -550,10 +551,11 @@ plt.legend()
 plt.show()
 ```
 
-Note: `f'$\\alpha = {α}$'` in the solution is an application of [f-String](https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings), which allows you to use `{}` to contain an expression. 
+```{note}
```
mmcky, line 554: @HumphreyYang I can't see this note anywhere in the output?

mmcky: @HumphreyYang so it turns out you can't seach for text in the dropdown items ... good to know.

Screen Shot 2022-09-21 at 7 50 19 pm

HumphreyYang: > @HumphreyYang I can't see this note anywhere in the output?

@mmcky I move it down to the solution part.

HumphreyYang: > @HumphreyYang so it turns out you can't seach for text in the dropdown items ... good to know.

I moved it down here because it is a bit too long and it does not fit into the structure of the hint if we add another note admonition. Would you like me to move it back to the hint?

jstac:
I think this looks great @HumphreyYang and @mmcky .

> I moved it down here because it is a bit too long and it does not fit into the structure of the hint if we add another note admonition.

Good choice, nice work.  Many thanks guys, merging.

PR #222, opened 2022-09-19
Adding Numpy Broadcasting

### LINKED ISSUE #206: Add discussion of broadcasting in NumPy
Add to NumPy lecture?

The images here are very helpful: https://jakevdp.github.io/PythonDataScienceHandbook/02.05-computation-on-arrays-broadcasting.html

We could make our own or link to that lecture.

Ideally, we would have

- the rules
- some images to provide intuition
- a discussion of how broadcasting can minimize loops
- an economic example

HumphreyYang:
Hi @jstac and @mmcky,

Thank you so much for sharing great resources on this topic.

I have added a section on broadcasting in Lecture 9. This PR resolves #206 .

Could you please kindly review this PR and let me know if there is anything I could improve?

Many thanks in advance.

### DIFF
--- a/lectures/numpy.md
+++ b/lectures/numpy.md
@@ -443,7 +443,7 @@ In particular, `A * B` is *not* the matrix product, it is an element-wise produc
 
 With all the element-wise operations above, what would happen if one of the matrix does not have corresponding elements.
 
-For example, `a` is a 3 by 3 matrix (`a -> (3, 3)`), but `b` is a 1 by 3 matrix (`b -> (1, 3)`).
+For example, `a` is a 3 by 3 array (`a -> (3, 3)`), but `b` is an array with three elements (`b -> (1, 3)`).
 
 Element-wise addition will result in a 3 by 3 matrix with each row adding `b`.
 
@@ -465,10 +465,6 @@ tags: [hide-input]
 
 import numpy as np
 from matplotlib import pyplot as plt
-
-# Draw a figure and axis with no boundary
-fig = plt.figure(figsize=(5, 1), facecolor='w')
-ax = plt.axes([0, 0, 1, 1], xticks=[], yticks=[], frameon=False)
 
 
 def draw_cube(ax, xy, size, depth=0.4,
@@ -531,6 +527,10 @@ dotted = dict(c='black', ls='-', lw=0.5, alpha=0.5,
               label_kwargs=dict(color='gray'))
 depth = 0.3
 
+# Draw a figure and axis with no boundary
+fig = plt.figure(figsize=(5, 1), facecolor='w')
+ax = plt.axes([0, 0, 1, 1], xticks=[], yticks=[], frameon=False)
+
 # first block
 draw_cube(ax, (1, 7.5), 1, depth, [1, 2, 3, 4, 5, 6, 9], '1', **solid)
 draw_cube(ax, (2, 7.5), 1, depth, [1, 2, 3, 6, 9], '2', **solid)
@@ -574,13 +574,15 @@ ax.text(5, 7.0, '+', size=12, ha='center', va='center')
 ax.text(10.5, 7.0, '=', size=12, ha='center', va='center');
 ```
 
-How about when `b -> (3, 1)`?
+How about `b -> (3, 1)`?
 
 ```{code-cell} python3
 b.shape = (3, 1)
 
 a + b
 ```
+
+We find that Numpy automatically fill up `b` from `b -> (3, 1)` to `b -> (3, 3)`
 
 ```{code-cell} python3
 ---
@@ -634,6 +636,7 @@ ax.text(10.5, 7.0, '=', size=12, ha='center', va='center');
 
 
 ```
+
 Element-wise addition will result in a (3, 3) matrix with each column adding `b`.
 
 
@@ -671,19 +674,63 @@ b = np.array([3, 6, 9])
 a + b
 ```
 
-The ValueError tells us that operands could not be broadcast together.
+The `ValueError` tells us that operands could not be broadcast together.
 
-Let's try to visualise why broadcasting only works in specific situations.
+```{code-cell} python3
+---
+tags: [hide-input]
+---
+# Draw a figure and axis with no boundary
+fig = plt.figure(figsize=(3, 1.3), facecolor='w')
+ax = plt.axes([0, 0, 1, 1], xticks=[], yticks=[], frameon=False)
 
-In the previous two examples where
- - `a -> (3, 3)`
- - `b -> (1, 3)` and `b -> (3, 1)`
+# first block
+draw_cube(ax, (1, 7.5), 1, depth, [1, 2, 3, 4, 5, 6, 9], '1', **solid)
+draw_cube(ax, (2, 7.5), 1, depth, [1, 2, 3, 6, 7, 9, 10], '2', **solid)
 
-we have additions like this:
+draw_cube(ax, (1, 6.5), 1, depth, [2, 3, 4], '4', **solid)
+draw_cube(ax, (2, 6.5), 1, depth, [2, 3, 7, 10], '6', **solid)
+
+draw_cube(ax, (1, 5.5), 1, depth, [2, 3, 4], '7', **solid)
+draw_cube(ax, (2, 5.5), 1, depth, [2, 3, 7, 10], '9', **solid)
+
+# second block
+draw_cube(ax, (6, 7.5), 1, depth, [1, 2, 3, 4, 5, 6, 9], '3', **solid)
+draw_cube(ax, (7, 7.5), 1, depth, [1, 2, 3, 6, 9], '6', **solid)
+draw_cube(ax, (8, 7.5), 1, depth, [1, 2, 3, 6, 7, 9, 10], '9', **solid)
+
+draw_cube(ax, (6, 6.5), 1, depth, range(2, 13), '3', **dotted)
+draw_cube(ax, (7, 6.5), 1, depth, [2, 3, 6, 7, 9, 10, 11], '6', **dotted)
+draw_cube(ax, (8, 6.5), 1, depth, [2, 3, 6, 7, 9, 10, 11], '9', **dotted)
+
+draw_cube(ax, (6, 5.5), 1, depth, [2, 3, 4, 7, 8, 10, 11, 12], '3', **dotted)
+draw_cube(ax, (7, 5.5), 1, depth, [2, 3, 7, 10, 11], '6', **dotted)
+draw_cube(ax, (8, 5.5), 1, depth, [2, 3, 7, 10, 11], '9', **dotted)
 
 
+ax.text(5, 7.0, '+', size=12, ha='center', va='center')
+ax.text(10, 7.0, '=', size=12, ha='center', va='center')
+ax.text(11, 7.0, '?', size=16, ha='center', va='center');
+```
 
+We can see that Numpy has difficulty in expanding this operation.
 
+It is because when `b` is expanded from `b -> (1, 3)` to `b -> (3, 3)`, Numpy is still having difficulties to match `b` with `a -> (3, 2)` without resulting in data loss. 
+
+Therefore, we can summerise a rule of thumb to check how Numpy will broadcast matrix:
+
+1. When the dimensions of two arrays do not match, Numpy will expand the one with less dimension by adding a dimension on the left of existing dimension.
+    <br> For example, `a -> (3, 3)` and `b -> (3)`, then broadcasting will add a dimension to the left of existing `b` so that `b -> (1, 3)`
+
+2. When the two arrays have the same dimension but different shapes, numpy will try to expand the smaller array to match the larger array.
+    <br> For example, `a -> (3, 3)` and `b -> (1, 3)`, then broadcasting will expand `b` so that `b -> (3, 3)`.
+    <br> We also have seen a situation where `a -> (3, 3)` and `b -> (3, 1)`, then broadcasting will also expand `b` so that `b -> (3, 3)`. 
+
+3. After step 1 and 2, the two arrays still do not match, a `ValueError` will be raised
+    For example, `a -> (3, 7)` and `b -> (3)`
+    <br> By step 1, b will be expanded to `b -> (1, 3)`;
+    <br> By step 2, b will be expanded to `b -> (3, 3)`.
+    <br> We can see that they do not match each other. Thus a `ValueError` will be raised.
 
 ### Matrix Multiplication
 
--- a/lectures/numpy.md
+++ b/lectures/numpy.md
@@ -443,13 +443,13 @@ In particular, `A * B` is *not* the matrix product, it is an element-wise produc
 
 In element-wise operations above, arrays may not have the same shape.
  
-Numpy will automatically expand arrays to the same shape whenever possible.
+NumPy will automatically expand arrays to the same shape whenever possible.
 
-This useful (but sometimes confusing) feature in Numpy is called **broadcasting**.
+This useful (but sometimes confusing) feature in NumPy is called **broadcasting**.
 
 For example, `a` is a 3 by 3 2-dimensional array (`a -> (3, 3)`), but `b` is an array with three elements (`b -> (3,)`).
 
-Numpy will automatically expand `b` from `b -> (3,)` to `b -> (3, 3)`.
+NumPy will automatically expand `b` from `b -> (3,)` to `b -> (3, 3)`.
 
 Element-wise addition will then result in a (3, 3) array
 
@@ -581,7 +581,7 @@ ax.text(10.5, 7.0, '=', size=12, ha='center', va='center');
 
 How about `b -> (3, 1)`?
 
-Numpy will automatically expand `b` from `b -> (3, 1)` to `b -> (3, 3)`
+NumPy will automatically expand `b` from `b -> (3, 1)` to `b -> (3, 3)`
 
 Element-wise addition will then also result in a (3, 3) matrix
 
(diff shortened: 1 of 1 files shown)


lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
 (numpy_matrix_multiplication)=
 In particular, `A * B` is *not* the matrix product, it is an element-wise product.
 
+(broadcasting)=
+#### Broadcasting
```
jstac: Looking at the flow of the lecture, this should be a level 3 heading, so please replace `####` with `###`

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+
+In element-wise operations above, arrays may not have the same shape.
+ 
+NumPy will automatically expand arrays to the same shape whenever possible.
+
+This useful (but sometimes confusing) feature in NumPy is called **broadcasting**.
+
+For example, `a` is a 3 by 3 2-dimensional array (`a -> (3, 3)`), but `b` is an array with three elements (`b -> (3,)`).
```
jstac: * For example, *suppose* 
* "but" should be "while" (after comma)
* "3 by 3"  should be $3 \times 3$
* `b` is *a flat* array

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+
+NumPy will automatically expand `b` from `b -> (3,)` to `b -> (3, 3)`.
```
jstac: When performing comparisons and arithmetic operations on `a` and `b`, NumPy will automatically...

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+
+Element-wise addition will then result in a (3, 3) array
```
jstac: How about changing this to "For example," ?

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+
+```{code-cell} python3
+
+a = np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]])
```
jstac:
```
a = np.array([[1, 2, 3], 
               [4, 5, 6], 
               [7, 8, 9]])
```

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+b = np.array([3, 6, 9])
+
+a + b
+```
+
```
jstac, line 520: "Here is a visual representation of this broadcasting operation:"

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+draw_cube(ax, (13, 5.5), 1, depth, [2, 3], '14', **solid)
+draw_cube(ax, (14, 5.5), 1, depth, [2, 3, 7, 10], '18', **solid)
+
+ax.text(5, 7.0, '+', size=12, ha='center', va='center')
+ax.text(10.5, 7.0, '=', size=12, ha='center', va='center');
+```
+
+How about `b -> (3, 1)`?
+
+NumPy will automatically expand `b` from `b -> (3, 1)` to `b -> (3, 3)`
```
jstac: In this case, NumPy will automatically...

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+
+Element-wise addition will then also result in a (3, 3) matrix
+
+```{code-cell} python3
+b.shape = (3,)
+
+a + b
+```
+
```
jstac, line 652: If you look at the deployment, it shows the wrong result.  Please be super careful not to introduce errors.

You need `b.shape = 3, 1`

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+draw_cube(ax, (14, 5.5), 1, depth, [2, 3, 7, 10], '18', **solid)
+
+ax.text(5, 7.0, '+', size=12, ha='center', va='center')
+ax.text(10.5, 7.0, '=', size=12, ha='center', va='center');
+
+
+```
+
+
+Broadcasting above can be seen as the following `for` loop
```
jstac: The previous broadcasting operation is equivalent to the following `for` loop

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+row, column = a.shape
+result = np.empty((3,3))
+for i in range(row):
+    for j in range(column):
+        result[i, j] = a[i, j] + b[i]
+
+print(result)
+```
+
+Broadcasting is actually efficient since we do not need to write `for` loops to implement element-wise operations when shapes do not agree.
```
jstac: How about "Broadcasting is more efficient than the `for` loop because we avoid the overhead of arithmentic operations in high-level Python code."   ?

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+
+In some cases, both operants will be expanded.
```
jstac: "operants" should be "operands"

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+
+When we have `a -> (3,)` and `b -> (3, 1)`
+
+```{code-cell} python3
+a = np.array([3, 6, 9])
+b = np.array([2, 3, 4])
+b.shape = (3,)
```
jstac: did you mean `b.shape = 3, 1` ?

Again, if you look in the deployment, your results are not correct.

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+draw_cube(ax, (12, 5.5), 1, depth, [2, 3, 4], '7', **solid)
+draw_cube(ax, (13, 5.5), 1, depth, [2, 3], '10', **solid)
+draw_cube(ax, (14, 5.5), 1, depth, [2, 3, 7, 10], '13', **solid)
+
+ax.text(5, 7.0, '+', size=12, ha='center', va='center')
+ax.text(10.5, 7.0, '=', size=12, ha='center', va='center');
+```
+
+Element-wise addition will result in a (3, 3) matrix with `a -> (3,)` being expanded to `a -> (3, 3)` and `b -> (3, 1)` being expanded to `b -> (3, 3)`.
+
```
jstac: ...results in a $3 \times 3$ matrix, with

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+It sounds great and straightforward, so why it is confusing?
```
jstac: While broadcasting is very useful, it can sometimes seem confusing.

For example, let's try...

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+
+
+ax.text(4.5, 7.0, '+', size=12, ha='center', va='center')
+ax.text(10, 7.0, '=', size=12, ha='center', va='center')
+ax.text(11, 7.0, '?', size=16, ha='center', va='center');
+```
+
+We can see that NumPy has difficulty expanding the arrays to the same size.
+
+It is because when `b` is expanded from `b -> (3,)` to `b -> (3, 3)`, NumPy is still having difficulties to match `b` with `a -> (3, 2)`.
```
jstac: "It is because when" -> "This is because, when"

"NumPy is still having difficulties to match" -> "NumPy cannot match"

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+
+It will get even trickier when we move to higher dimensions.
```
jstac: "It will" -> "Things get"

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+
+Fortunately, we have a list of rules for broadcasting thanks to [Jake VanderPlas](https://jakevdp.github.io/PythonDataScienceHandbook/02.05-computation-on-arrays-broadcasting.html):
```
jstac: How about "To help us, we can use the following list of rules:"

(Your version suggests that Jake VanderPlas made these rules, or was the first one to write them down.  But I think they will be documented somewhere in NumPy.)

lectures/numpy.md
jstac: Please add here

"(This section draws on an excellent discussion of broadcasting provided by [Jake VanderPlas](https://jakevdp.github.io/PythonDataScienceHandbook/02.05-computation-on-arrays-broadcasting.html).)"

Or maybe replace "draws on" with "summarizes" ?

lectures/numpy.md
```diff
@@ -438,6 +438,409 @@ A * B
...
+    - When `a -> (2, 2, 2)` and `b -> (2, 2)`, then broadcasting will add a dimension to the left so that `b -> (1, 2, 2)`.
+    - When `a -> (3, 2, 2)` and `b -> (2)`, then broadcasting will add a dimension to the left so that `b -> (1, 1, 2)`.
+
+
+* Step 2: When the two arrays have the same dimension but different shapes, NumPy will try to expand arrays to match each other.
+    - For example, when `a -> (1, 3)` and `b -> (3, 1)`, then broadcasting will expand both `a` and `b` so that `a -> (3, 3)` and `b -> (3, 3)`;
+    - When `a -> (2, 2, 2)` and  `b -> (1, 2, 2)`, then broadcasting will expand `b` so that `b -> (2, 2, 2)`. 
+    - When `a -> (3, 2, 2)` and `b -> (2,)`, then broadcasting will add a dimension to the left so that `b -> (1, 1, 2)`.
+
+Here are code examples for boardcasting higher dimensional arrays
```
jstac: "boardcasting" -> "broadcasting"

lectures/numpy.md
```diff
@@ -985,3 +1388,40 @@ F.plot(ax)
...
 ```{solution-end}
 ```
+
+
+```{exercise-start}
+:label: np_ex4
+```
+
+Recall [broadcasting](broadcasting) in Numpy can help us conduct element-wise operations between arrays with different number of dimensions without using `for` loop.
+
```
jstac, line 1450: Recall *that*

"between" -> "on"

"loop" -> "loops"

lectures/numpy.md
```diff
@@ -985,3 +1388,40 @@ F.plot(ax)
...
+In this exercise, try to use `for` loop to replicate the result of the following code
```
jstac: use *a* `for` loop

jstac:
@HumphreyYang Many thanks --- although it would be good if this was a bit closer to ready for merge by the time I looked at it.

Many comments are minor / suggestions, but some concern errors.

I wonder if we need a system for another set of eyes to pass over it (CC @mmcky) ?

mmcky:
@HumphreyYang for big changes like this -- why don't you tag me as a reviewer first and then after my review we can pass this over to @jstac for final review.

HumphreyYang:
> @HumphreyYang for big changes like this -- why don't you tag me as a reviewer first and then after my review we can pass this over to @jstac for final review.

Thank you so much for the advice, I will follow this workflow for the next PR.

HumphreyYang:
> @HumphreyYang Many thanks --- although it would be good if this was a bit closer to ready for merge by the time I looked at it.
> 
> Many comments are minor / suggestions, but some concern errors.
> 
> I wonder if we need a system for another set of eyes to pass over it (CC @mmcky) ?

Thanks for your detailed review. The error was introduced when I change the notation for dimensions in the text after checking the code. I apologise for missing these apparent typos. I will fix these errors and send them to @mmcky before the final version is submitted to you.

HumphreyYang:
Hi @mmcky, 

Thanks for your review of other PRs. I will integrate your suggestions as soon as possible.

I have made several updates to this PR to reflect previous reviews. Could you also please also review the latest deployment and let me know if anything needs to be improved?

Many thanks.

mmcky:
Thanks @HumphreyYang having some additional details of `numpy` broadcasting will be a great addition to the lecture. 

@jstac my main thinking on this is whether it is in the right place. `Broadcasting` is a pretty tricky subject and a to of users will use `numpy` (as infrastructure) in the early phases to do array operations. So my thinking would be to:

1. remove `Operations on Arrays` section and then have
2. promote `Arithmetic Operations` to H2
3. then include `Matrix Multiplication` to H2
4. then include `Broadcasting` section as H2

in my mind this enables users to get up and running with arrays quickly and then extends their knowledge into broadcasting rules which is hugely important but somewhat tricky. I think promoting these to H2 also enables them to be found more easily on the right hand Contents.

HumphreyYang:
> Thanks @HumphreyYang having some additional details of `numpy` broadcasting will be a great addition to the lecture.

Hi @mmcky, thanks for the feedback, are there any other tricky implementations of broadcasting I should add to the current version?

jstac:
+1 @mmcky , good calls.  I agree.

Perhaps it's worth adding a sentence at the start of the section on broadcasting as follows.

Note: Some aspects of broadcasting are hard to visualize and considered relatively advanced.  As such this section can be skimmed at first pass.

mmcky:
> > Thanks @HumphreyYang having some additional details of `numpy` broadcasting will be a great addition to the lecture.
> 
> Hi @mmcky, thanks for the feedback, are there any other tricky implementations of broadcasting I should add to the current version?

Sorry for confusion @HumphreyYang -- what I was saying is having `numpy` broadcasting will be a nice addition to the lecture.

HumphreyYang:
Hi @mmcky,

I have changed the structure to reflect your suggestions. It looks great to me. Would you mind checking the latest deployment to see if there is anything I could improve when you are available?

Many thanks!

mmcky:
thanks @HumphreyYang I think this looks good and we can tag @jstac for final approval.

mmcky (approved):

lectures/numpy.md
```diff
@@ -475,6 +478,456 @@ A @ (0, 1)
...
+row, column = a.shape
+result = np.empty((3, 3))
+for i in range(row):
+    for j in range(column):
+        result[i, j] = a[i, j] + b[i]
+
+result
+```
+
+Broadcasting is more efficient than the `for` loop because we avoid the overhead of arithmetic operations in high-level Python code.
```
jstac: This sentence should be moved to the start of the section, so people have motivation to read up on broadcasting.  In fact I suggest we change it to

The value of broadcasting is that

* `for` loops can be avoided, which helps numerical code run fast and
* broadcasting can allow us to implement operations on arrays without actually creating some dimensions of these arrays in memory, which can be important when arrays are large.

lectures/numpy.md
```diff
@@ -475,6 +478,456 @@ A @ (0, 1)
 
 Since we are post-multiplying, the tuple is treated as a column vector.
 
+(broadcasting)=
+## Broadcasting
+
+```{index} single: NumPy; Broadcasting
+```
+
+(This section is built upon an excellent discussion of broadcasting provided by [Jake VanderPlas](https://jakevdp.github.io/PythonDataScienceHandbook/02.05-computation-on-arrays-broadcasting.html).)
```
jstac: On second thoughts, please change "is built upon" to "extends".

@HumphreyYang , you have done some very nice work that extends the discussion by Jake VanderPlas.

lectures/numpy.md
```diff
@@ -475,6 +478,456 @@ A @ (0, 1)
...
+
+b = np.array(
+    [[1,7], 
+     [7,1]])
+print(f'the shape of array b is {b.shape}')
+
+a + b
+```
+
 ### Mutability and Copying Arrays
```
jstac: Currently this section "Mutability and Copying Arrays" appears as a subsection of "Broadcasting", which is not correct.

I suggest that we replace `###` here with `##`.

lectures/numpy.md
```diff
@@ -475,6 +478,456 @@ A @ (0, 1)
...
+## Broadcasting
+
+```{index} single: NumPy; Broadcasting
+```
+
+(This section is built upon an excellent discussion of broadcasting provided by [Jake VanderPlas](https://jakevdp.github.io/PythonDataScienceHandbook/02.05-computation-on-arrays-broadcasting.html).)
+
+```{note}
+Some aspects of broadcasting are hard to visualize and considered relatively advanced. As such this section can be skimmed at first pass.
+```
```
jstac, line 491: On second thoughts, let's change the wording of this note to:

Broadcasting is a very important aspect of NumPy.  At the same time, advanced broadcasting is relatively complex and some of the details below can be skimmed on first pass.

lectures/numpy.md
```diff
@@ -475,6 +478,456 @@ A @ (0, 1)
...
+We can see that NumPy cannot expand the arrays to the same size.
+
+It is because, when `b` is expanded from `b -> (3,)` to `b -> (3, 3)`, NumPy cannot match `b` with `a -> (3, 2)`.
+
+Things get even trickier when we move to higher dimensions.
+
+To help us, we can use the following list of rules:
+
+* *Step 1:* When the dimensions of two arrays do not match, NumPy will expand the one with fewer dimensions by adding dimension(s) on the left of the existing dimensions.
+    - For example, suppose `a -> (3, 3)` and `b -> (3,)`, broadcasting will add a dimension to the left so that `b -> (1, 3)`;
```
jstac: Here and below, please change the English syntax to "if ABC, then XYZ".  For example

"suppose `a -> (3, 3)` and `b -> (3,)`, broadcasting" should be "if `a -> (3, 3)` and `b -> (3,)`, then broadcasting"

jstac:
This is very good @HumphreyYang.  It's great to have this discussion of broadcasting added.

Broadcasting is very important for Google JAX, so it's increasingly relevant for scientific Python.

Could you please make the small changes requested and then ask @mmcky to merge.

@mmcky Many thanks for reviewing.  Please merge and publish when ready.  Could you please ping @natashawatkins once this is published so she can tweet this too.

HumphreyYang:
Hi @mmcky,

I have incorporated comments from @jstac into the latest deployment. Could you please take a quick look at it and merge it if it looks good to you?

Many thanks.

mmcky:
thanks @HumphreyYang for all your work on this. Thanks @jstac for review.

PR #235, opened 2022-10-01
Adding Performance Evaluation to Numpy Exercise 

### LINKED ISSUE #230: Add speed comparison for broadcasting exercise in NumPy lecture
I saw that there was a new exercise in the NumPy lecture ([Exercise 9.4](https://python-programming.quantecon.org/numpy.html#exercises)), but I thought maybe it would be more useful for illustrating broadcasting if the reader is asked to use compare the speeds of broadcasting vs. using the for-loop they write.

HumphreyYang:
Hi @mmcky,

This resolves #230. Could you please kindly review this PR and let me know if there is anything I could change?

Many thanks

### DIFF
--- a/lectures/numpy.md
+++ b/lectures/numpy.md
@@ -1451,9 +1451,24 @@ Recall that [broadcasting](broadcasting) in Numpy can help us conduct element-wi
 In this exercise, try to use a `for` loop to replicate the result of the following code
 
 ```{code-cell} python3
-x = np.random.randn(4, 4)
-y = np.random.randn(4)
-x / y
+import quantecon as qe
+
+np.random.seed(123)
+x = np.random.randn(1000, 100, 100)
+y = np.random.randn(100)
+qe.tic()
+A = x / y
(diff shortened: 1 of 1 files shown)


mmcky:
thanks @HumphreyYang I think this is a good addition.

mmcky:
@jstac if you approve then I'll merge.

jstac:
Thanks @HumphreyYang , nice work.

My only comment is that in #230 I suggested "adding a second part to the exercise where 4 is replaced by a large number and the reader is asked to time a for loop against the broadcast implementation."

In other words, the reader first replicates on a small example to make sure that they are getting the same result, and then tests the two versions on a larger example to investigate timing.

Do you think this is worthwhile?

HumphreyYang:
Hi @jstac, 

I have added back the sample example. Could you have another look at the newest deployment?

Many thanks.

jstac:
Perfect.  Thanks @HumphreyYang!  Merging.

PR #237, opened 2022-10-03
LECTURE: An Introduction to JAX

jstac:
Closes https://github.com/QuantEcon/lecture-python-programming.myst/issues/224

This is a first pass at adding a JAX lecture.  It needs a very good proofread -- I wrote it quickly! -- and it probably needs to be running on a GPU to make sense.

Looking forward to seeing this come together!  All help is appreciated @mmcky @HumphreyYang @Smit-create .

PR #239, opened 2022-10-04
ENH: Enable Custom Runner using CML for GPU Support (PREVIEWS)

mmcky:
This PR enables building on a custom runner hosted on EC2 using `p2.xlarge` instances for PREVIEW builds

These instances have:

1. 1 x NVIDIA K80 GPU
1. 4 x vCPU's

All setup and use of EC2 is automated and works the same way as the previous Preview Workflow Script.

The `cml` runner manages the EC2 instance so the instance only runs when preview builds are running. As a result it takes a little longer to setup as it fetches a fresh instance each time a preview build runs. It then uses a Docker image based on `ubuntu` and installs anaconda etc.  However this means we are only paying for EC2 when previews are getting built and the instance is torn down when the preview run is completed. This makes this option very cost effective. 

Checks

- [x] it will process each commit as an independent run

https://github.com/QuantEcon/meta/issues/56


mmcky:
@jstac I propose I enable the following custom build runner workflow (which builds on `EC2`) for previews only and do some test run on the next few PR's, before we look at publishing from these runs.

So `GPU` support is available currently for:

- [x] previews
- [ ] publishing

The publishing will need to be enabled in a subsequent PR. 

**Note:** Test build cache workflows to ensure that works the same way as GitHub Actions hosted workflow.

mmcky:
@HumphreyYang I will merge this PR soon which will change how `previews` are build via `GitHub Actions`. It would be great if you could take close looks at previews for the next couple of changes and let me know if you can spot any issues. 

This PR is bringing the ability to build `code` that requires access to a `GPU` so we can test this out on #237

HumphreyYang:
> @HumphreyYang I will merge this PR soon which will change how `previews` are build via `GitHub Actions`. It would be great if you could take close looks at previews for the next couple of changes and let me know if you can spot any issues.
> 
> This PR is bringing the ability to build `code` that requires access to a `GPU` so we can test this out on #237

Thanks @mmcky. I will keep an eye on the new previews.

PR #243, opened 2022-11-01
ENH: GPU Support for Preview, Cache and Publish

mmcky:
This PR adds GPU support for:

- [x] cache
- [x] preview
- [x] publish

build workflows as they form the `publishing` pipeline.

To reduce costs we will continue to build:

- execution-linux, execution-win, execution-osx 
- link checker

on `GitHub actions` with `cpu` only support.

### DIFF
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -4,11 +4,35 @@ on:
     branches:
       - main
 jobs:
-  tests:
+  deploy-runner:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
+      - uses: iterative/setup-cml@v1
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Deploy runner on EC2
+        env:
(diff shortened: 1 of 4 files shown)


mmcky:
@HumphreyYang the hardware we have settled on is:

| Name | Value |
| ------ | ------|
| GPU | Tesla V100 |
| # of vCPU | 8 | 
| vCPU | Broadwell E5-2686 v4 |


with the following software for the GPU

| Name | Value |
| ------ | ------|
| Nvidia driver | 470.141.03 |
| CUDA version | 11.4 |

mmcky:
@HumphreyYang would you be able to take a final look at this preview and let me know if you spot any issues with rendering etc. It should all checkout but it would be nice as a double check just prior to changing our publishing workflows. I will also need to confirm that the gh-pages action works on a remote ec2 instance.

mmcky:
@HumphreyYang I am going to merge this PR -- once I confirm the `publish` workflow has worked I will ask if you could have a quick review of the live site.

HumphreyYang:
Hi @mmcky, 

I did not spot any abnormality in the preview. I will have another look at the live site once deployed. 

Thanks!

Issue #247, opened 2022-11-01
Intro with statement


jstac:
We should probably introduce the with statement in https://python-programming.quantecon.org/python_essentials.html#input-and-output

Reference: https://www.geeksforgeeks.org/with-statement-in-python/

PR #253, opened 2022-11-21
[pthon_essentials] Adding context managers (with statements)

HumphreyYang:
Hi @mmcky,

This PR addresses #247. Could you have a quick look at the additions when the preview is ready and merge if it looks good to you?

Many thanks in advance.

### DIFF
--- a/lectures/python_essentials.md
+++ b/lectures/python_essentials.md
@@ -326,31 +326,40 @@ out
 print(out)
 ```
 
-We can also use `with` statement to contain operations on the file within a block for clarity and cleanness.
+We can also use a `with` statement (also known as a [context manager](https://realpython.com/python-with-statement/#the-with-statement-approach)) which enables you to group operations on the file within a block which improves the clarity of your code.
+
 
 ```{code-cell} python3
-with open('newfile.txt', 'a+') as file: 
-
-    # Adding a line at the end 
-    file.write('\nAdding a new line')
-
-    # Move the pointer back to the top
-    file.seek(0)
-
-    # Read the file line-by-line from the first line
+with open("newfile.txt", "r") as f:
+    file = f.readlines()
+with open("output.txt", "w") as fo:
     for i, line in enumerate(file):
-        print(f'Line {i}: {line}')
+        fo.write(f'Line {i}: {line} \n')
 ```
 
-Note that we used `a+` mode (standing for append+ mode) to allow appending new content at the end of the file and enable reading at the same time.
+```{code-cell} python3
+fo = open('output.txt', 'r')
+out = fo.read()
+print(out)
+```
 
-There are [many modes](https://www.geeksforgeeks.org/reading-writing-text-files-python/) you could set when opening the file.
+We can write the example above under the same context to reduce redundancy
 
-When experimenting with different modes, it is important to check where the pointer is set.
+```{code-cell} python3
+with open("newfile.txt", "r") as f, open("output2.txt", "w") as fo:
+        for i, line in enumerate(f):
+            fo.write(f'Line {i}: {line} \n')
+```
 
-The pointer is set to the end of the file in `a+` mode.
+```{code-cell} python3
+fo = open('output2.txt', 'r')
+out = fo.read()
+print(out)
+```
 
-Here we set `file.seek(0)` to move the pointer back to the first line (line 0) of the file to print the whole file.
+```{note}
+Note that we only used `r` and `w` mode. There are [more modes](https://www.geeksforgeeks.org/reading-writing-text-files-python/) you could experiment with.
+```
 
 
 ### Paths
--- a/lectures/python_essentials.md
+++ b/lectures/python_essentials.md
@@ -326,8 +326,7 @@ out
 print(out)
 ```
 
-We can also use a `with` statement to ensure the files are properly acquired 
-and released.
+In fact, the recommended approach in modern Python is to use a `with` statement to ensure the files are properly acquired and released.
 
 Containing the operations within the same block also improves the clarity of your code.
 
--- a/lectures/python_essentials.md
+++ b/lectures/python_essentials.md
@@ -326,26 +326,32 @@ out
 print(out)
 ```
 
-We can also use `with` statement to contain operations on the file within a block.
-
-Note that we used `a+` mode (standing for append+ mode) to allow appending new content at the end of the file and enable reading at the same time.
-
-There are even [more modes](https://www.geeksforgeeks.org/reading-writing-text-files-python/) you could set when openning the file.
-
-The trick is to check where the pointer is set: if the pointer is set at the end of the file `seek` method is needed to go back to the start of the file. 
-
-Here we set `file.seek(0)` to move the pointer back to the first line (line 0) of the file.
+We can also use `with` statement to contain operations on the file within a block for clarity and cleanness.
 
 ```{code-cell} python3
 with open('newfile.txt', 'a+') as file: 
+
+    # Adding a line at the end 
     file.write('\nAdding a new line')
+
+    # Move the pointer back to the top
     file.seek(0)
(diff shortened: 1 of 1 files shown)


lectures/python_essentials.md
```diff
@@ -326,6 +326,33 @@ out
 print(out)
 ```
 
+We can also use `with` statement to contain operations on the file within a block for clarity and cleanness.
```
mmcky:
```suggestion
We can also use a `with` statement (also known as a [context manager](https://realpython.com/python-with-statement/#the-with-statement-approach)) which enables you to group operations on the file within a block which improves the clarity of your code.
```

lectures/python_essentials.md
```diff
@@ -326,6 +326,33 @@ out
...
+
+    # Move the pointer back to the top
+    file.seek(0)
+
+    # Read the file line-by-line from the first line
+    for i, line in enumerate(file):
+        print(f'Line {i}: {line}')
+```
+
+Note that we used `a+` mode (standing for append+ mode) to allow appending new content at the end of the file and enable reading at the same time.
```
mmcky:
````suggestion
```{note} 
We used `a+` mode (standing for append+ mode) to allow appending new content at the end of the file and enable reading at the same time.
```
````

**Review by mmcky — COMMENTED** [review]
@HumphreyYang just made a few suggestions. 

My only other comment is that I can't remember a time whereI have ever written and read from a file in the same `with` statement before. I just wonder if this is overly complicating the example. I typically use `a+` to append to an already established file. Thoughts?

**Comment by HumphreyYang** [comment]
> @HumphreyYang just made a few suggestions.
> 
> My only other comment is that I can't remember a time when I have ever written and read from a file in the same `with` statement before. I just wonder if this is overly complicating the example. I typically use `a+` to append to an already established file. Thoughts?

Hi @mmcky, 

Thanks for your detailed review. I was overcomplicating the example to show that there are other modes that are capable of doing operations other than `r` and `w` modes. I think the standard practice is always to have a `with` statement to write and a `with` statement to read. Just curious if that is what you would suggest (i.e. breaking this down into two `with` statements).

**Comment by mmcky** [comment]
> > @HumphreyYang just made a few suggestions.
> > My only other comment is that I can't remember a time when I have ever written and read from a file in the same `with` statement before. I just wonder if this is overly complicating the example. I typically use `a+` to append to an already established file. Thoughts?
> 
> Hi @mmcky,
> 
> Thanks for your detailed review. I was overcomplicating the example to show that there are other modes that are capable of doing operations other than `r` and `w` modes. I think the standard practice is always to have a `with` statement to write and a `with` statement to read. Just curious if that is what you would suggest (i.e. breaking this down into two `with` statements).

thanks @HumphreyYang I would usually go down that path -- but I am a conservative programmer. I think though this would be good practice.

**Comment by jstac** [comment]
Please let me know when to review @mmcky and @HumphreyYang .  

We can keep this addition to the lectures very simple and straightforward.  Our only aim is to get people up to speed so that they can code well enough to handle subsequent econ lectures.

**Comment by HumphreyYang** [comment]
Hi @mmcky,

I have pushed a new version with the section on the pointer removed entirely since it will inevitably lead to complicated examples with many `seek` and `tell` printings.

**Comment by HumphreyYang** [comment]
> Please let me know when to review @mmcky and @HumphreyYang .
> 
> We can keep this addition to the lectures very simple and straightforward. Our only aim is to get people up to speed so that they can code well enough to handle subsequent econ lectures.

Thanks for the feedback @jstac, I will adjust the style to fit into this objective.

File: lectures/python_essentials.md [hunk 3]
```diff
@@ -326,6 +326,42 @@ out
 print(out)
 ```
 
+We can also use a `with` statement (also known as a [context manager](https://realpython.com/python-with-statement/#the-with-statement-approach)) which enables you to group operations on the file within a block which improves the clarity of your code.
+
+
+```{code-cell} python3
+with open("newfile.txt", "r") as f:
+    file = f.readlines()
+with open("output.txt", "w") as fo:
```
mmcky: @HumphreyYang should block be indented to be in the open file context?

mmcky (commented):
thanks @HumphreyYang it is simpler. 

Just had one question re: indentation

HumphreyYang:
> thanks @HumphreyYang it is simpler.
> 
> Just had one question re: indentation

Thanks for picking that out @mmcky. I have added an indentation to the block. Should we invite John for review if it looks good to you?

HumphreyYang:
> thanks @HumphreyYang it is simpler.
> 
> Just had one question re: indentation

Thanks for picking that out @mmcky. I have added an indentation to the block. Should we invite John for review if it looks good to you?

jstac:
Hi @mmcky , it says above "This branch has not been deployed. No deployments."  I can't see anything related to netlify.

Am I missing something?

mmcky:
The last `box` in the checks section of the PR (need to scroll down as its the last item in the list) contains the Netlify link to the latest `green tick` deployment. 

The Netlify integration used to post a link each run but not sure why that behaviour changed.

https://6383fedc52896355884dbec5--epic-agnesi-957267.netlify.app/intro.html

That `not deployed` message (as I understand) is if we used `netlify` to actually host he site it would confirm the deployment status. It is a bit misleading.

jstac:
> need to scroll down as its the last item in the list

Ah, that's where I was going wrong.  To be fair, it's a bad idea to have the downward scroll indicator hidden until you hover...

Thanks @mmcky  :-)

lectures/python_essentials.md
jstac: This sentence is too long and uses "which" twice.

Please remember that there must always be a comma before which.  ("...within a block, which...")

"We can also use..." is the start of the previous sentence in the file so it's better not to reuse it.

jstac:
Hi guys, I thought the point of the `with` statement is: "The with statement itself ensures proper acquisition and release of resources."  See https://www.geeksforgeeks.org/with-statement-in-python/

Perhaps this should be mentioned directly, so people can see the point.

Also, I recommend that we start with the simple example

```python
f = open('newfile.txt', 'r')
out = f.read()
out
```

and show how that changes when `with` is used.  It's usually best to start with the simplest example possible.

Then, after that, you could say: "Now suppose that we want to read input from one file and write output to another.  Here's how we could do accomplish this task while correctly acquiring and returning resources to the operating system:"

lectures/python_essentials.md
```diff
@@ -326,6 +326,42 @@ out
...
+
+```{code-cell} python3
+with open("newfile.txt", "r") as f:
+    file = f.readlines()
+    with open("output.txt", "w") as fo:
+        for i, line in enumerate(file):
+            fo.write(f'Line {i}: {line} \n')
+```
+
+```{code-cell} python3
```
jstac, line 371: Shouldn't we use `with` here, given that we just recommended it?

lectures/python_essentials.md
```diff
@@ -326,6 +326,42 @@ out
...
+
+We can write the example above under the same context to reduce redundancy
+
+```{code-cell} python3
+with open("newfile.txt", "r") as f, open("output2.txt", "w") as fo:
+        for i, line in enumerate(f):
+            fo.write(f'Line {i}: {line} \n')
+```
+
+```{code-cell} python3
```
jstac, line 386: Shouldn't we use `with` here, as above?

jstac:
Great work @HumphreyYang .  This is very nice.

Just this comment to attend to: 

> "We can also use..." is the start of the previous sentence in the file so it's better not to reuse it.

How about "In fact, the recommended approach in modern Python is to use a `with` statement..."

HumphreyYang:
> Member

Thanks for the feedback @jstac. It does sound much better. I have updated it accordingly.

Many thanks.

jstac:
Thanks again @HumphreyYang , very nice examples.  Merging.

PR #254, opened 2022-11-27
Edited GPU note

jstac:


lectures/jax_intro.md
```diff
@@ -13,20 +13,27 @@ kernelspec:
 
 # JAX
 
+
+```{note}
+This lecture is built using [hardware](status:machine-details) that
+has access to a GPU. This means that 
+
+1. the lecture might be significantly slower when running on your machine, and
+2. the code is well-suited to execution with Google colab (click "Launch Notebook" button).
```
jstac: @mmcky , can this be replaced by a link?  That is, "...execution with [Google colab](link to launch notebook on colab)."

PR #255, opened 2022-11-28
ENH: Build GPU badge to identify lectures that use GPU hardware

mmcky:
This PR looks at using a linked badge to identify lectures that use `gpu`

- [x] can badges be added to titles?
- [ ] Issue with LaTeX support due to unicode characters

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,26 +44,26 @@ jobs:
         run: |
           pip install --upgrade "jax[cuda]" -f https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
           nvidia-smi
-      - name: Install latex dependencies
-        shell: bash -l {0}
-        run: |
-          apt-get -qq update
-          export DEBIAN_FRONTEND=noninteractive
-          apt-get install -y tzdata
-          apt-get install -y     \
-            texlive-latex-recommended \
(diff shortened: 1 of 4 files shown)


mmcky:
- [ ] update this when https://github.com/executablebooks/jupyter-book/pull/1842 is released

mmcky:
@jstac we could add badges to lectures such as the screenshot below that are clickable to a page that talks about the GPU setup now that we have upgraded to the latest `jupyter-book`

Screenshot 2023-04-07 at 9 14 31 am

jstac:
Sounds good @mmcky .  

Note that this lecture is flagged to be shifted by you to the new series https://github.com/QuantEcon/lecture-jax/issues/1  

(There are a couple of lectures with JAX that won't be shifted --- the ones created by Tom where the focus is less on JAX and more on an economic application.)

mmcky:
@jstac for sure. This was just a test case as a demo. 

> Note that this lecture is flagged to be shifted by you to the new series https://github.com/QuantEcon/lecture-jax/issues/1

I am waiting until the new `jax` series get's a bit closer to going public to reduce the possible editing / update issues. I had planned to migrate these in the week we plan for the series to go live. Does that make sense?

jstac:
Thanks @mmcky , that makes a lot of sense :-)

PR #258, opened 2022-12-05
[More Language Features] Adding * Operators

### LINKED ISSUE #257: [More Language Features] Add Splat Operator, *args, and **kwargs to Lecture 17
Adding a new section on  Splat Operator, `*args`, and `**kwargs` in lecture 17.

HumphreyYang:
This PR resolves #257 by adding a section on the use of `*`.

### DIFF
--- a/lectures/python_advanced_features.md
+++ b/lectures/python_advanced_features.md
@@ -314,7 +314,7 @@ Now we can use `*` to unpack the lists
 
 ```{code-cell} python3
 l1 = ["a", "b", "c", "d", "e"]
-l2 = ["e", "f"]
+l2 = ["e", "f", "j"]
 l3 = [*l1, *l2]
 
 print(l3)
@@ -368,17 +368,23 @@ def arb(*args):
 arb(l1, l2)
 ```
 
+Let's try more inputs
+
+```{code-cell} python3
+arb(l1, l2, l3, l4)
+```
+
 We can see that `*args` allows us to put an arbitary number of input into the function, and the inputs are stored into a tuple.
 
 Now we can use it together with what we have learnt before to write compute the union of multiple lists
 
 ```{code-cell} python3
 def intersect(*lists):
-    output = []
-    for ls in lists:
-        print(f'list is {ls}')
-        output = [*ls, *output]
-    return set(output)
+    output = set()
+    for i in range(len(lists)):
+        other = set([j for l in (lists[:i] + lists[i+1:]) for j in l])
+        output.update(set(lists[i]).intersection(other))
+    return output
 ```
 
 ```{code-cell} python3
@@ -392,8 +398,51 @@ intersect(l1, l2, l3, l4)
 
 We find that the function is now capable of handling arbitarily many lists inputs.
 
-Suppose we also want to compute unique elements
+Suppose we also want to output intersections for each list. In this case, we will need to input keyword arguments with an arbitary size.
 
+Similarly, Python allows us to use `**kargs` to pass keyword arguments into functions without specifying the size of the arguments.
+
+Let's start with a simple example
+
+```{code-cell} python3
+def arb(**args):
+    print(args)
+
+arb(l1=l1, l2=l2)
+```
+
+We can see that Python uses a dictionary to store these keyword arguments.
+
+Let's try more inputs
+
+```{code-cell} python3
+arb(l1=l1, l2=l2, l3=l3, l4=l4)
+```
+
+Now suppose we want the intersection function to return both intersection and unique elements in each list in a dictionary.
+
+We can use `**kargs` to pass names of the lists as well lists into the fucntion
+
+
+```{code-cell} python3
+def intersect(**lists):
+    keys = list(lists.keys())
+    print(keys)
+    inters_output = set()
+    unique_output = dict()
+    for i in range(len(keys)):
+        key = keys[i]
+        other = set([j for k in (keys[:i] + keys[i+1:]) for j in lists[k]])
+        inters = set(lists[key]).intersection(other)
+        inters_output.update(inters)
+        unique_output[key] = set(lists[key]) - inters
+    return inters_output, unique_output
+```
+
+```{code-cell} python3
+intersectionSet, uniqueSet = intersect(l1=l1, l2=l2, l3=l3, l4=l4)
+print(intersectionSet, uniqueSet)
+```
 
 
 
--- a/lectures/python_advanced_features.md
+++ b/lectures/python_advanced_features.md
@@ -352,14 +352,14 @@ for book in book_list:
     printBook(**book)
 ```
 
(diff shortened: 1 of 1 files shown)


HumphreyYang:
Hi @mmcky,

Could you have a quick review on this PR to see where I should further improve?

Many thanks in advance!

mmcky:
@HumphreyYang I have updated the `build` infrastructure from the `main` branch. I would like to use this change as a final test case. Let me know if you spot any issues. Thanks.

HumphreyYang:
> @HumphreyYang I have updated the `build` infrastructure from the `main` branch. I would like to use this change as a final test case. Let me know if you spot any issues. Thanks.

Many thanks @mmcky, the build runs much faster! I did not see any notable issues in the build.

lectures/python_advanced_features.md
```diff
@@ -290,6 +290,227 @@ tags: [raises-exception]
...
+
+When we have a very long list of structured dictionaries, this can will make our function scalable.
+
+
+Overall, when `*list`/`*tuple` and `**dictionary` are passed into *function calls*, they will be unpacked into individual arguments instead of a collection.
+
+The difference is that `*` will unpack lists and tuples the into positional arguments, while `**` will unpack dictionaries into keyword arguments.
+
+
+### Arbitrary Arguments
```
mmcky, line 407: This section seems a bit `arbitrary` :-). I almost never use this outside of writing code for a library so I guess I am wondering if this should be included here. @HumphreyYang do you have any good examples of where this is used in `analysis` and `research` style code?

mmcky (commented):
@HumphreyYang thanks for putting this together. 

I am not sure that we should talk much about arbitrary functions `*args and **kwargs` as I don't think this is a great fit for the intended audience of this lecture. What do you think?

lectures/python_advanced_features.md
```diff
@@ -290,6 +290,227 @@ tags: [raises-exception]
 max(y)
 ```
 
+## `*` and `**` Operators
+
+`*` and `**` are convenient and widely used tools to unpack lists and tuples and to allow users to define functions that use arbitrarily many arguments as input.
```
mmcky: TBH - I almost never use these operators in my code as I find them one aspect of `python` syntax that is not memorable :-). I think @jstac uses these operators when passing parameter sets between models as a use case in the economics / research analysis domain. Can we think of some possible examples that do some things in this context?

HumphreyYang: Hi @mmcky, many thanks for your feedback. I think you are absolutely right about the use cases. I proposed writing something about this because it is also widely used in AI/ML packages and research papers (given we also have readers who are interested in data science :) ). I think my original intention is also to introduce these concepts to readers so that they can recognise and understand them; especially when they read the documentation and source code of scientific packages later, but it does feel like I have overdone some of the sections.

I think I will reorganise this section and include small functions in `qe` package and explain what they are used for in the context of the function, making it more practical, and reorienting the focus to understanding them instead of writing them. 

Please kindly let me know if this sounds better to you.

jstac: I think this is a good plan @HumphreyYang .  What you have written is good and clear but would be even better if the examples were a little closer to research.

In this lecture there's a plotting example where I unpack dictionaries: https://python.quantecon.org/lqcontrol.html
(See, e.g., `axes[1].plot(list(range(T+1)), assets, 'b-', label="assets", **p_args)`)

lectures/python_advanced_features.md
```diff
@@ -290,6 +290,227 @@ tags: [raises-exception]
...
+print(l3)
+```
+
+The output is what we are expecting -- a list with elements in `l1` and `l2`.
+
+```{note}
+It works in this case because `[]` can be considered a function call.
+```
+
+However, we cannot use `*` in list comprehensions and outside function calls.
```
jstac: Even though this is a good point @HumphreyYang , and well written, I suggest that we drop it to minimize content, remembering that we always want to keep our lectures as short as possible.  We shouldn't allocate too much space to this topic.

HumphreyYang:
Hi @mmcky,

Thanks for your spot-on feedback last time. I have pushed a simplified version of the PR. I made some choices during the process: 

- I have used a conceptually easier example instead of the [one based on GBM](https://63a2c1f7fd31a44bbbc01a37--epic-agnesi-957267.netlify.app/) I wrote before.

- I also left out an exercise that asks readers to replicate the `generate_plots` function in the example using `args` and `**kargs`. The solution will be something like this:

```
def generate_plots(*β_ls, idx=0, **kargs):
    label_list = []
    for βs in zip(β_ls[0::2], β_ls[1::2]):
        ax[idx].plot(*generate_data(*βs), **kargs['line_kargs'])
        label_list.append(f'$β0 = {βs[0]}$ / $β1 = {βs[1]}$')
    ax[idx].legend(label_list, **kargs['legend_kargs'])

β_ls = [10, 1,
        20, 2,
        30, 3]

generate_plots(*β_ls, line_kargs=line_kargs, legend_kargs=legend_kargs, idx=0)

β_ls.extend([40, 1])

generate_plots(*β_ls, line_kargs=line_kargs, legend_kargs=legend_kargs, idx=1)

plt.show()
```
I think this exercise is a good one to help readers to distinguish `*` `**` in the context of function definitions and function calls, but I left out this part because it uses assumed knowledge from many previous lectures and may misdirect readers. 

I was curious about what is your opinions on the current version and the compromises I have made.

Many thanks in advance. Have a lovely Christmas and New Year! 🎄

mmcky:
thanks @HumphreyYang I have taken another look at the simplified version. I am still wondering if this is `too much` -- it is a good example though. I might be really biased as I tend not to use these operators much at all. 

@jstac do you have good example? I imagine the most common `quantecon` use case is when passing parameters to models?

lectures/python_advanced_features.md
```diff
@@ -290,6 +290,180 @@ tags: [raises-exception]
...
+Let's walk through a simple example together and distinguish the use of `*` and `**`
+
+```{code-cell} python3
+import numpy as np
+import matplotlib.pyplot as plt
+
+# Set up the frame and subplots
+fig, ax = plt.subplots(2, 1)
+plt.subplots_adjust(hspace=0.7)
+plt.rcParams["figure.figsize"] = (10, 6)
```
jstac: Do we need line 356?  This is usually at the top of each lecture.

jstac:
Hi all,

I think this is maybe a fraction long, given that we work hard to keep discussion very short, but nicely written and a good example.  I recommend merging conditional on addressing the minor comment above.

Nice work @HumphreyYang .

HumphreyYang:
Hi @mmcky,

Thanks for your reviews. I have addressed the issue @jstac raised. Could you please merge this PR if it looks to you?

Many thanks.

mmcky:
thanks @HumphreyYang and @jstac will merge this in now.

PR #269, opened 2023-04-04
ENH: Upgrade anaconda==2023.03 + CUDA 12.1

mmcky:
This PR

- updates `anaconda==2023.03`
- uses `cuda 12.1` with `jaxlib==0.4.8`

### DIFF
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -27,18 +27,16 @@ jobs:
     needs: deploy-runner
     runs-on: [self-hosted, cml-gpu]
     container:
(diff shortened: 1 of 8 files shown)


mmcky:
@HumphreyYang @Smit-create if you have time can you please take a quick look through this test build as we are upgrading the anaconda distribution and moving to latest version of cuda and jax

Smit-create:
I don't see any broken things but it would be nice to have a code-cell printing the JAX backend in all the JAX lectures so that we can easily spot bugs related to GPU/CPU maintenance.

PR #273, opened 2023-04-13
FIX: GA for latest qe-theme

AakashGfude:
updating the config variable for google analytics

### DIFF
--- a/lectures/_config.yml
+++ b/lectures/_config.yml
@@ -2,8 +2,6 @@ title: Python Programming for Economics and Finance
 author: Thomas J. Sargent & John Stachurski
 logo: _static/qe-logo.png
 description: This website presents a set of lectures on python programming for economics, designed and written by Thomas J. Sargent and John Stachurski.
-analytics:
-  google_analytics_id: UA-54984338-9
 
 execute:
   execute_notebooks: "cache"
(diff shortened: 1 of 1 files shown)


lectures/_config.yml
```diff
@@ -32,6 +30,8 @@ sphinx:
       og_logo_url: https://assets.quantecon.org/img/qe-og-logo.png
       description: This website presents a set of lectures on python programming for economics, designed and written by Thomas J. Sargent and John Stachurski.
       keywords: Python, QuantEcon, Quantitative Economics, Economics, Sloan, Alfred P. Sloan Foundation, Tom J. Sargent, John Stachurski
+      analytics:
```
mmcky, line 33: hey @AakashGfude is this specific to `quantecon-book-theme` as `sphinx-book-theme` moved it to the top level?

AakashGfude: @mmcky it is because of pydata-sphinx-theme. From what I read in their latest code: https://github.com/pydata/pydata-sphinx-theme/blob/main/docs/user_guide/analytics.rst . Did `sbt` do that in their code? 
Locally, the ga code is getting included when I do this.

mmcky: interesting -- ok thanks @AakashGfude

mmcky:
@DrDrij I will merge this as a test case using `lecture-programming` and we can see if the correct `html` is included for the `G4` codes.

DrDrij:
> @DrDrij I will merge this as a test case using `lecture-programming` and we can see if the correct `html` is included for the `G4` codes.

Test case looks to work well @mmcky. I've setup a GA4 code PR here: https://github.com/QuantEcon/lecture-python-programming.myst/pull/274

PR #292, opened 2023-10-04
MAINT: Upgrade software stack

mmcky:
This PR upgrades to:

- anaconda==2023.09
- python=3.11
- quantecon-book-theme==0.6.0

### DIFF
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -14,7 +14,7 @@ jobs:
           auto-update-conda: true
           auto-activate-base: true
           miniconda-version: 'latest'
-          python-version: "3.10"
+          python-version: "3.11"
           environment-file: environment.yml
           activate-environment: quantecon
       - name: Build HTML
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,13 +33,13 @@ jobs:
       - name: Display Pip Versions
         shell: bash -l {0}
         run: pip list
-      - name: Download "build" folder (cache)
-        uses: dawidd6/action-download-artifact@v2
-        with:
-          workflow: cache.yml
-          branch: main
-          name: build-cache
-          path: _build
+      # - name: Download "build" folder (cache)
+      #   uses: dawidd6/action-download-artifact@v2
+      #   with:
+      #     workflow: cache.yml
+      #     branch: main
+      #     name: build-cache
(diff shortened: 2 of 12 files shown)


mmcky:
@HumphreyYang would you have time to look into this link checker result?

mmcky:
- [x] re-run without cache

HumphreyYang:
Many thanks @mmcky,

The link checker is all green now.

mmcky:
thanks @HumphreyYang love your work.

mmcky:
- [ ] re-enable the build `cache`

mmcky:
@HumphreyYang @HengchengZhang if you had 10minutes would you mind to look through the latest deployment and let me know if you spot any issues or regressions.

HengchengZhang:
Hi Matt, the new search bar and the dark mode looks all good to me. 

One minor thing I can mention is that the color of code block in dark mode, they are now the same color as in bright mode with a dark background. I think it would be better to have a seperate color map for that if possible? 

It is still great if we use the current theme.

mmcky:
> Hi Matt, the new search bar and the dark mode looks all good to me.
> 
> One minor thing I can mention is that the color of code block in dark mode, they are now the same color as in bright mode with a dark background. I think it would be better to have a seperate color map for that if possible?
> 
> It is still great if we use the current theme.

Thanks @HengchengZhang -- I agree we need to make improvements to the theme. Are you indicating that `quantecon-book-theme==0.6.0` is worse than our current live site which uses `quantecon-book-theme==0.5.3`?

HengchengZhang:
> > Hi Matt, the new search bar and the dark mode looks all good to me.
> > One minor thing I can mention is that the color of code block in dark mode, they are now the same color as in bright mode with a dark background. I think it would be better to have a seperate color map for that if possible?
> > It is still great if we use the current theme.
> 
> Thanks @HengchengZhang -- I agree we need to make improvements to the theme. Are you indicating that `quantecon-book-theme==0.6.0` is worse than our current live site which uses `quantecon-book-theme==0.5.3`?

Hi @mmcky I think the new `quantecon-book-theme==0.6.0` is better than current site but the code part can be improved. But still the new theme is pretty nice.

mmcky:
thanks @HumphreyYang and @HengchengZhang for your review and comments. 

Upgrading to the latest versions now.

PR #301, opened 2023-12-20
MAINT: upgrade quantecon-book-theme==0.7.1

mmcky:


mmcky:
- [x] linkchecker results are false positives so will merge

PR #328, opened 2024-11-04
TST: build html first

mmcky:
This PR is a test for diagnosing issues with the current cache `_build` folder mechanism.


mmcky:
No longer required

PR #329, opened 2024-11-04
FIX: enable .jupyter_cache to be included in cache runs

mmcky:
This PR fixes a bug that has prevented the cache mechanism to work. 

Hidden files were not included by default in the upload-artifact action

https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#hidden-files

### DIFF
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -32,3 +32,4 @@ jobs:
         with:
           name: build-cache
           path: _build
+          include-hidden-files: true


kp992:
Oh, this looks great. Nice catch!

But I am wondering why it wasn't failing 2-3 months ago

kp992:
I see, it is because of this: https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#hidden-files. The version change of Action file is the issue. Thanks @mmcky

PR #347, opened 2025-06-04
Update getting_started.md

jaeyungkim:


jstac:
Thanks @jaeyungkim .  

@mmcky , could you please review?

mmcky:
thanks @jaeyungkim 

the command

```
conda update conda
```

will actually update the conda manager tool.

Perhaps we should update this to be

```
conda update conda
conda update anaconda
```

what do you think?

mmcky:
It actually looks like this behaviour has changed in `conda` and no longer possible. 

I have been doing some local testing and with an older conda environment it no longer works as it used to. 

It seems conda is moving towards being just a package manager and being less focused on a set of tools that work well together with each release.

jstac:
Thanks @mmcky .  What is no longer possible?

Is there an action you can recommend to resolve this issue?

mmcky:
> Thanks @mmcky . What is no longer possible?
> 
> Is there an action you can recommend to resolve this issue?

```
conda update anaconda
```

> Is there an action you can recommend to resolve this issue?

I am reviewing it now. Doing some testing. We will then update `getting_started.md` for sure. 

@jstac The easiest way to upgrade anaconda is a fresh install, but hesitate to recommend that in getting_started just yet :-)

mmcky:
@jstac also

```
conda update --all
```

works just fine but it will update all packages to the latest version, not the latest anaconda distribution. 

That might be the "best" recommendation though? However this will then create environments that deviate from our testing regime. 

I have been checking `anaconda` releases and the last one was `anaconda=2024.10` -- they have really slowed down the release schedule.

jstac:
Perhaps we should just remove that line or provide a link to the conda docs...

mmcky:
thanks @jstac 

the suggestion by @jaeyungkim is great.

What was confusing is the section title is `Updating Anaconda` so I have now changed that to Updating conda and that then matches the content of that section well. Updating conda will always work well.

mmcky (approved):

PR #354, opened 2025-06-10
Fix inconsistency between text and code in Section 5.4.1 - format() vs f-strings

jaeyungkim:
Issue
The explanatory text in Section 5.4.1 mentions using `format()` method, but the actual code uses f-strings. This might create confusion for learners.

PS)
The tutorial is excellent overall - this is just a small clarification to improve the learning experience.


jstac:
Thanks @jaeyungkim , these corrections are much appreciated.

@mmcky , would you mind taking a look?

mmcky:
thanks @jaeyungkim I have updated the lecture in #355

PR #356, opened 2025-06-13
Update python_oop.md

jaeyungkim:
"instance data and functions" together are called attributes, not methods. I believe methods are just the function part.

### DIFF
--- a/lectures/python_oop.md
+++ b/lectures/python_oop.md
@@ -192,7 +192,7 @@ In our example, an *instance* will be the name of  particular *person* whose *in
 
 In our example, two functions `earn` and `spend` can be applied to the current instance data.
 
-Taken together,  the instance data and functions  are called *methods*.
+Taken together,  the instance data and functions  are called *attributes*.
 
 These can be readily accessed in ways that we shall describe now.
 


jstac:
Thanks @jaeyungkim .

This is indeed completely wrong and we appreciate the fix!

jstac:
@mmcky Could you please merge when ready and perhaps consider republishing?

mmcky:
thanks @jstac and @jaeyungkim. 

I thought that (to be exact) in Python classes

```python
self.example_variable = {{ value }}
```

is an `attribute`

while 

```python
def function(self, args, kwargs)
    {{ function code }}
```

is a `method`. 

So classes have both `attributes` being variables and data and `methods` being functions. 

We could be more precise with the statement for sure (thanks @jaeyungkim) so in my view there is still an error in changing to attributes. I'll propose an update.

lectures/python_oop.md
```diff
@@ -192,7 +192,7 @@ In our example, an *instance* will be the name of  particular *person* whose *in
 
 In our example, two functions `earn` and `spend` can be applied to the current instance data.
 
-Taken together,  the instance data and functions  are called *methods*.
+Taken together,  the instance data and functions  are called *attributes*.
```
mmcky, line 195:
```suggestion
In a Class,  the instance data are called *attributes* and functions are called *methods*.
```

mmcky: There are also `properties` but I don't think we need to mention those.

jstac:
People probably have different opinions here.  Google Gemini says the following:

Yes, in Python, a method is an attribute of a class.

To be precise, a method is a callable attribute of a class. Let's break down what this means:

jstac:
Please feel free to make the call @mmcky , I'll leave it with you.

mmcky:
Thanks @jstac -- interesting. In that case `attributes` is definitely an improvement. I'll merge this and make live.

PR #345, opened 2025-05-23
[Experimental] Use mystmd to build lectures

kp992:
Update the lectures and build with the new `mystmd`.

PR #363, opened 2025-07-23
[Experimental] Build the lectures with QuantEcon/mystmd instead of jupyter-book

kp992:
Experimental setup for building the lectures with the QuantEcon MyST stack. This PR is built on top of #345.

## Current toolchain

Both workflows now build with the QuantEcon fork of the MyST CLI and the new theme, so the Netlify preview shows the latest content rendered by the latest tooling.

| Piece | What the branch uses | Where it is set |
| --- | --- | --- |
| MyST CLI | [`QuantEcon/mystmd`](https://github.com/QuantEcon/mystmd), cloned and built from source with bun, but **the two workflows differ**: `ci.yml` is pinned to the exact SHA `96ee78f5` so failures stay bisectable, while `build-ipynb.yml` clones `--branch main` and so tracks the fork's latest. `quantecon/VERSION.yml` is printed to the log, so every build records the `qe-vN` state it ran with — the ipynb build is currently on `qe-v10`, the HTML build on the pinned June engine | `.github/workflows/ci.yml` (pinned), `.github/workflows/build-ipynb.yml` (`main`) |
| Theme | [`quantecon-theme.mystmd`](https://github.com/QuantEcon/quantecon-theme.mystmd) `v3.0.0`, pinned to the release zip (taken in #629) | `lectures/myst.yml` (`site.template`) |
| Build command | `myst build --html --execute` | `.github/workflows/ci.yml` |
| Content | merged up to date with `main` | — |

The theme moves off the archived `QuantEcon/quantecon-theme` bundle (which the branch was still pulling as a floating `main.zip`) onto a pinned release of the new theme repo. Bump the `vX.Y.Z` in the `site.template` URL to take a newer theme; the fork stays on `main` deliberately, so the preview tracks the latest tooling.

## The build was not actually using the fork

Worth calling out separately, because it had been green and invisible: until this update the HTML preview was built by `jupyter book build`, which **cannot** use the fork. jupyter-book 2.x is a thin Python shim — `jupyter_book/__main__.py` execs a compiled mystmd bundle vendored inside its own wheel (`jupyter_book/dist/jupyter-book.cjs`) and exposes no hook to point it at another CLI. The globally installed fork was simply ignored, and only the `build-ipynb` workflow (which calls `myst` directly) ever exercised it.

The last run before the fix shows it plainly: the jupyter-book HTML build emitted 27 × `'output' unknown export output extension: exports/<page>.ipynb`, because the per-page ipynb export is a fork feature (`myst-to-ipynb`, `qe-v1`) that upstream does not know about. The same commit built by the fork's `myst build --ipynb` emitted none and exported 23 clean notebooks. The HTML build now runs `myst build --html --execute` — the same engine and the same flags, since jupyter-book 2.x *is* mystmd, just our build of it.

## Everything else in this update

- **Merged `main`.** Brings in the new Polars lecture, the numba rewrite, the `pandas_panel` switch to remote CSVs, and the translation-sync workflows. The one conflict was `ci.yml`, where `main` and this branch have rewritten the same job for different builders — resolved in favour of this branch's job, keeping `main`'s runner spec and action bumps.
- **`myst.yml` TOC re-synced with `_toc.yml`.** It had drifted: `autodiff` and the new `polars` lecture were missing, `workspace` appeared twice, and one part title differed. The two files now list exactly the same 27 pages.
- **`project.github` corrected** to `lecture-python-programming` — it pointed at `lecture-python-programming.myst`, which now only survives as a rename redirect. At the time this mattered because the theme derived the notebook-launch repo from the value plus a `.notebooks` suffix; **theme v3.0.0 removed that derivation**, so `project.github` now only drives the commit and "edit this page" links, and the launch repository is named explicitly (see #629).
- **JAX is now installed in the build.** Previous preview builds were rendering `ModuleNotFoundError` cells for `jax_intro`, `autodiff` and `numpy_vs_numba_vs_jax`, because this job installed only `myst_requirements.txt` while the jupyter-book 1.x job on `main` installs `jax[cuda13]` separately. Same install, same runner family.
- **CI actions moved to current majors and Node 20 → 24** (the theme's `.nvmrc`; it runs as a Remix server during `myst build --html`). This also clears the Node 20 runtime deprecation warnings. The unused `configure-pages` step is gone — this job deploys to Netlify, not Pages.
- **Removed committed execution artifacts** (`foo.py`, `newfile.txt`, `numbers.txt`, `output.txt`, `output2.txt`, `test_table.csv`, `us_cities.txt`). All are written by `%%file` / `%%writefile` cells in the lectures themselves, all are matched by `.gitignore`, and none exist on `main`.

## Still open

- **Nothing fails the build on a notebook execution error.** Execution errors are reported and the build carries on, which is how the JAX failures above survived green builds. `main` gained a gate for the jupyter-book 1.x path in #588; the fork's CLI has `myst build --strict` for this, but it should be turned on only once the remaining build diagnostics are triaged, or it will fail on unrelated warnings.
- ~~Launch-button URLs are left on the theme defaults.~~ **Resolved by #629.** Theme v3.0.0 makes Launch opt-in, so `site.options` now names `launch_notebook_repo`, `launch_notebook_source_dir` and `launch_colab`, ported from `lectures/_config.yml` rather than invented; the branch is left unset because `main` is the default. Confirmed on that PR's preview: the button resolves to `…/lecture-python-programming.notebooks/blob/main/need_for_speed.ipynb`.
- **The `ci.yml` fork pin's comment is out of date.** It says the SHA is "the exact SHA of the fork's main (unmoved since 2026-06-12)", but the fork's `main` has since moved to `c1f274b8a` (2026-09-10) and several `qe-vN` checkpoints have landed. The pin itself is a deliberate choice — it keeps failures bisectable — but the HTML build is now running a materially older engine than `build-ipynb.yml`, which is worth a conscious decision rather than drift.


mmcky:
@kp992 thanks for opening this. Excellent work! It is exciting to see how close we are now. 

From my review it looks like there a just a few issues and features we need before launch. 

**Bugs:**

1. LHS TOC menu flickers when using the Home page or a top level Part section. It is like it has an auto-stow feature that is triggering when pages are reloaded. It doesn't look great when pages are refreshed. 
2. Code block formatting has too much indentation for Solution blocks

**Features:**

1. I see a bunch of new `export` notebooks being added to the `yaml` headers of each lecture. Do we need to do some development work on supporting the download notebook link in the theme? 
3. Is `index` support on the jb2 / mystmd roadmap?

mmcky:
## CommonMark ipynb Export Support

This commit adds **CommonMark notebook export** to the JB2 branch, enabling `myst build --ipynb` to produce notebooks with plain CommonMark markdown cells — compatible with vanilla Jupyter Notebook, JupyterLab (without `jupyterlab-myst`), and Google Colab.

### Changes

**24 lecture files** — added `markdown: commonmark` to each export config:

```yaml
exports:
  - format: ipynb
    markdown: commonmark      # new
    output: exports/<name>.ipynb
```

**New CI workflow** (`.github/workflows/build-ipynb.yml`) — runs in parallel with the existing JB2 workflow:

- Clones [QuantEcon/mystmd@myst-to-ipynb](https://github.com/QuantEcon/mystmd/tree/myst-to-ipynb) and builds from source
- Runs `myst build --html` and `myst build --ipynb` using the custom CLI
- Audits all 24 exported notebooks for MyST syntax leaks
- Uploads ipynb exports as build artifacts

### How it works

The `markdown: commonmark` option triggers an AST pre-transform in `myst-to-ipynb` that converts MyST-specific nodes to their CommonMark equivalents before serialization:

| MyST syntax | CommonMark output |
|---|---|
| `:::{note}` admonitions | `> **Note**` blockquotes |
| Math roles | `$E=mc^2$` dollar signs |
| Math blocks | `$$...$$` dollar-dollar |
| `:::{figure}` / `{image}` | `alt` |
| `{exercise}` / `{solution}` | **Exercise N** / **Solution** blocks |
| `{proof}` / theorem | **Theorem (Title)** blocks |
| `(label)=` targets | Stripped |
| `{index}` directives | Stripped |
| `+++` block markers | Stripped |

### Local test results

All 24 notebooks exported successfully with **0 MyST syntax leaks** (verified by automated audit script).

### Dependencies

This uses the [myst-to-ipynb branch on QuantEcon/mystmd](https://github.com/QuantEcon/mystmd/tree/myst-to-ipynb) which contains 5 commits:

- `79c7be0b` — Bug fixes (kernelspec, language_info, log message, block markers)
- `f67f1188` — CommonMark serialization mode (~475 lines)
- `c1cca05f` — Test suite (35 tests across 3 YAML files)
- `2d70076d` — Real-world validation fixes (identifier stripping, empty cell filter)
- `2c7fa5d7` — Image directive handler (found during full-project testing)

mmcky:
@kp992 I have been working on the `myst-to-ipynb` feature. 

Here are a collection of test notebooks: https://github.com/QuantEcon/lecture-python-programming.myst/actions/runs/22381335617 (can download the artifact).

This is looking pretty good I think.

kp992:
Thanks @mmcky for pushing this forward and fixing the issues. I saw the generated notebooks and they look great! One issue I found was about the solution rendering. The solution shows up as the normal markdown ```python code block instead of an executable block.

mmcky:
> Thanks @mmcky for pushing this forward and fixing the issues. I saw the generated notebooks and they look great! One issue I found was about the solution rendering. The solution shows up as the normal markdown ```python code block instead of an executable block.

Thanks @kp992 would you mind raising an issue on https://github.com/QuantEcon/mystmd

mmcky:
> > Thanks @mmcky for pushing this forward and fixing the issues. I saw the generated notebooks and they look great! One issue I found was about the solution rendering. The solution shows up as the normal markdown ```python code block instead of an executable block.
> 
> Thanks @kp992 would you mind raising an issue on https://github.com/QuantEcon/mystmd

Thanks @kp992 I think I have identified the root cause of this. https://github.com/QuantEcon/mystmd/issues/5
Working on a fix.

kp992:
Thank you @mmcky. The issue looks resolved and the generated notebooks are awesome!!

mmcky:
- [ ] check if theme regression is occurring. Clicking on a chapter briefly shows a malformed webpage that disappears to the LEFT. NEEDS REVIEW.

mmcky:
## Refreshed against the current toolchain

Branch is up to date with `main` and now builds with the QuantEcon fork of the MyST CLI and the new theme repo. The PR description has been rewritten to describe the resulting setup.

| Piece | Before | Now |
| --- | --- | --- |
| MyST CLI (HTML build) | upstream mystmd, vendored inside the `jupyter-book` wheel | [`QuantEcon/mystmd`](https://github.com/QuantEcon/mystmd) `main` — `v1.10.1 (qe-v8)` |
| Theme | `QuantEcon/quantecon-theme` floating `main.zip` (repo archived) | [`quantecon-theme.mystmd`](https://github.com/QuantEcon/quantecon-theme.mystmd) `v2.2.0`, pinned release zip |
| Content | 2 months behind `main` | merged, conflict resolved — PR is mergeable again |

### The HTML build was never using the fork

This is the part worth reading. `jupyter book build` **cannot** use our fork, and had been quietly building the preview with upstream mystmd this whole time. jupyter-book 2.x is a thin Python shim: `jupyter_book/__main__.py` execs a compiled mystmd bundle vendored inside its own wheel (`jupyter_book/dist/jupyter-book.cjs`) and exposes no hook to point it at a different CLI. The `npm install -g` of the fork in this workflow was installing a CLI that the build step then ignored. Only `build-ipynb.yml`, which calls `myst` directly, was ever exercising the fork.

The evidence is in the last run before the fix — the jupyter-book HTML build emitted 27 × `'output' unknown export output extension: exports/<page>.ipynb`, since the per-page ipynb export is a fork feature (`myst-to-ipynb`, `qe-v1`) that upstream doesn't know about. The same commit built by the fork's `myst build --ipynb` emitted none and exported 23 clean notebooks.

The HTML build now runs `myst build --html --execute` with the fork's CLI, and the `pip install jupyter-book` step is gone — it was only putting a second, conflicting mystmd on the image. Same engine and same flags either way; jupyter-book 2.x *is* mystmd, this is just our build of it.

### Three other things the review turned up

- **The JAX lectures were rendering `ModuleNotFoundError` cells.** This job installed only `myst_requirements.txt`, while the jupyter-book 1.x job on `main` installs `jax[cuda13]` separately, so `jax_intro`, `autodiff` and `numpy_vs_numba_vs_jax` had been executing to import errors through green builds. Same install added here.
- **`myst.yml`'s TOC had drifted from `_toc.yml`** — `autodiff` and the new `polars` lecture missing, `workspace` listed twice, one part title stale. The two files now list exactly the same 27 pages.
- **`project.github` pointed at `lecture-python-programming.myst`**, which only survives as a rename redirect. The theme derives the notebook-launch repo from that value plus a `.notebooks` suffix, so a stale name means broken launch buttons. Corrected.

Also: Node 20 → 24 (the theme's `.nvmrc`), remaining actions to current majors (clears the Node 20 removal deprecation), the unused `configure-pages` step dropped since this job deploys to Netlify rather than Pages, and seven committed execution artifacts removed (`foo.py`, `us_cities.txt`, `test_table.csv` and friends — all written by `%%file` cells in the lectures, all gitignored, none present on `main`).

### Left for a decision

Nothing fails the build on a notebook execution error — that is exactly how the JAX breakage stayed invisible. `main` gained a gate for the jupyter-book 1.x path in #588; the fork's CLI has `myst build --strict` for the same job here, but it is worth turning on only after the remaining build diagnostics are triaged, otherwise it will fail on unrelated warnings.

mmcky:
This is supporting @DrDrij review of quantecon-theme.mystmd

mmcky:
## Why this branch builds with QuantEcon/mystmd rather than jupyter-book

Retitled to reflect what the build chain actually does. The short version: **jupyter-book cannot be pointed at our fork**, so building through it silently discards every QuantEcon feature we depend on.

### jupyter-book ignores the fork, even when the fork is installed

jupyter-book 2.x is a Python shim that execs a mystmd bundle vendored inside its own wheel (`jupyter_book/dist/jupyter-book.cjs`). Installing `QuantEcon/mystmd` globally with npm afterwards puts our `myst` on `PATH`, but jupyter-book never calls it.

This is not a theoretical concern — it is visible in our own CI history. In the last run that built with `jupyter book build` (`4314bc19`, 2026-07-31), the `Verify mystmd version` step reported `v1.10.1 (qe-v8)`, confirming the fork was installed and first on `PATH`. The build in the very next step then emitted **23 errors** of the form `⛔️ unknown export output extension: exports/<lecture>.ipynb` — one per lecture declaring an ipynb export. That export format is a fork-only capability, so those errors prove the build was running upstream mystmd, not ours. Under `myst build` with the fork, the same 23 exports produce **zero** errors.

The `Verify mystmd version` step is therefore misleading on its own: it verifies the standalone binary, not the engine jupyter-book actually execs. It reports a reassuring `qe-v8` while the build beside it uses something else.

### What we would lose

The fork is at `qe-v8` on `upstream_base: 1.9.1` and carries thirteen QuantEcon features, most of them load-bearing for a book-style render: CommonMark ipynb export with image attachments, book-style numbering, Roman-numbered part dividers with a continuous chapter counter, section-level scope for `proof:*` / figure / equation numbering, the full title→heading chain per section, shared counters across proof kinds, Pandoc-style fancy ordered lists, citation scanning that respects inline code spans, `numbering.code` honouring the book prefix, and amsthm-convention unnumbered proofs. None of these exist upstream, so none of them survive a `jupyter book build`.

### What is and is not migrated

| workflow | engine |
|---|---|
| `ci.yml` (PR preview) | `myst build --html --execute`, QuantEcon/mystmd |
| `build-ipynb.yml` | `myst build --html` / `--ipynb`, QuantEcon/mystmd |
| `publish.yml` (production deploy) | still jupyter-book 1.x `jb build` — untouched from `main` |
| `cache.yml`, `execution-*.yml`, `environment.yml` | still jupyter-book 1.x — untouched from `main` |

Migrating the production chain is deliberately **out of scope here** and belongs in its own PR: it covers the PDF build (currently `jb build --builder pdflatex`, with a typst `book-pdf` export stubbed in `myst.yml`), the notebook downloads (currently the sphinx-tojupyter custom builder, already prototyped as `myst build --ipynb` in `build-ipynb.yml`), the weekly execution cache, and the Anaconda execution tests.

### Known issue on this branch

Cold-cache builds currently fail. The execution cache key is `hashFiles('lectures/**/*.md')`, so editing any single lecture invalidates all ~25 notebooks and forces a full re-execution; `jax_intro.md`'s kernel is then killed (`Connection lost` → `Kernel: restarting` → `Unhandled error`, exit 1) with no Python traceback.

This is not caused by mixing build systems — `ci.yml` installs no jupyter-book at all, and `publish.yml` / `cache.yml` do not run on pull requests. Comparing the last green cold run against the failures: same JAX 0.11.0, same 25 notebooks executed cold, same peak concurrency of 8, near-identical overlapping peers. The only difference is the engine — `jax_intro.md` executed in 25 s under jupyter-book's vendored mystmd and dies at 66 s under the fork's CLI. No timeout explanation survives either, since `numpy.md` took 99 s in the green run and built fine.

Two things worth doing before diagnosing further: pin the fork clone to the `qe-v8` tag rather than `--branch main`, since the engine is currently a moving target between runs of the same commit; and capture `free -m`, `nvidia-smi` and `dmesg -T | grep -i "killed process"` in an `if: always()` step, because a SIGKILL with no traceback is consistent with the host OOM killer but that remains unconfirmed.

mmcky:
## Cold-cache execution failure: diagnosed, fixed, validated

Closing out the "known issue" from my earlier comment. The kernel death was **not** the engine and **not** host memory — it was a GPU preallocation race, confirmed by an instrumented run and fixed in two commits.

### Diagnosis (instrumented run 30806359596)

A 5-second GPU/host-memory sampler ran beside the failing build. The GPU timeline told the whole story on one line: **0 → 11,305 MiB → 14,727 → 3,427 → 0** on the g4dn's 15,360 MiB T4.

The three JAX lectures (`jax_intro`, `autodiff`, `numpy_vs_numba_vs_jax`) execute concurrently (myst parallelism = cpus−1 = 7). The first JAX kernel to touch the GPU preallocated JAX's default 75% — 11,305 MiB. `jax_intro`'s kernel lost that race, clamped to the ~3.4 GiB remainder, and aborted ~40 s later when a cell outgrew the pool — an XLA fatal in the kernel process, which is why there was no Python traceback. dmesg was clean: no OOM-killer records, no segfaults, so the host is exonerated. The only kernel event in the entire build was `jax_intro`'s; the 11.3 GiB holder's exit was a clean post-build reap.

This also explains the confusing history: the jupyter-book-era cold run went green because `jax_intro` **won** the race that day (verified from its HTML artifact — JAX cos on 50M elements in 0.0022 s vs NumPy's 1.07 s, i.e. genuinely on GPU), and `main` never hits this because the JB1 build executes notebooks sequentially. The engine swap changed execution *phasing*, not semantics — which is why the failure appeared engine-correlated for three runs straight.

### Fix and validation

| commit | change |
| --- | --- |
| `8af1015` | instrumentation + engine pinned to the fork's exact SHA (`96ee78f`) instead of `--branch main` |
| `94a7e52` | `XLA_PYTHON_CLIENT_PREALLOCATE: 'false'` on the build step — concurrent kernels allocate on demand |
| `f541ccc` | hardening: cache key now hashes `myst_requirements.txt` + `ci.yml` alongside the lectures, JAX pinned to 0.11.0 |

Validation run 30806924581: fully cold (all 25 notebooks executed), zero kernel restarts, `jax_intro` built in 46 s (vs 45 s in the healthy jupyter-book-era cold run), peak GPU usage 4,217 MiB with no preallocation spike, dmesg clean. The `f541ccc` run is a second cold validation by construction, since changing the cache key forces one.

### Structural fixes, so this class of bug can't ship green again

The engine swap originally shipped green because the execution cache key hashed only `lectures/**/*.md` — a warm cache replayed old outputs and the new engine never executed a notebook. The key now hashes the requirements and the workflow file itself, so any engine SHA bump, JAX bump, or XLA env change forces a cold, genuinely-executed build automatically. The memory sampler and dmesg read-back stay in the workflow as standing telemetry.

One upstream follow-up filed from this: myst reports a dead kernel as a bare `Unhandled error` with no notebook, cell, cause, or kernel stderr — QuantEcon/mystmd#77 asks for the diagnostics that would have made this a one-look failure instead of an instrumented investigation.

kp992:
Thanks @mmcky for pushing this forward. Please let me know if you need help in taking this to finish line.

mmcky:
thanks @kp992 this is currently updated and built to help us get our theme for mystmd across the line. @DrDrij is doing a pass on that now -- so once that is in place we will then be in a position to migrate.

DrDrij:
Netlify preview only, not a production concern: assets here are served with `cache-control: public,max-age=0,must-revalidate`, so every navigation makes a blocking revalidation round-trip (~390ms for a 304 with 0 bytes) before first paint. GitHub Pages serves `max-age=600`, so the deployed lectures are unaffected. If it is worth making the preview behave like production, a `_headers` file with `/build/_assets/*  Cache-Control: public,max-age=31536000,immutable` is safe — those filenames are content-hashed, so a rebuild changes the URL.

mmcky:
@DrDrij I have just updated this to the latest release of quantecon-theme.mystmd incorporate recent merges.

mmcky:
Bumped the theme pin in `lectures/myst.yml` from v2.3.0 to **v2.5.0** (7a392b8), released today. Between those tags the theme picked up the static-build asset URL fixes (2.3.1), lecture content typography and the inline-code backticks regression fix (2.4.0), and in 2.5.0 the icon-only search trigger, PT Serif headings on the Sphinx scale, QuantEcon-blue content links, the Colab-only launcher (the Private JupyterHub option is gone), and the back-to-top / FOUC first-paint fixes. Release notes: https://github.com/QuantEcon/quantecon-theme.mystmd/releases/tag/v2.5.0. No other change on the branch; the existing conflict with `main` is untouched.

mmcky:
Bumped the pinned theme to [v2.6.0](https://github.com/QuantEcon/quantecon-theme.mystmd/releases/tag/v2.6.0) in b9524ff9. It brings the QuantEcon code-token palette and seoul256 text colours (Phase 3), and the language switcher, hreflang alternates, RTL support and translator credit (Phases 4–5). This edition sets none of the new multilingual options, so the only visible change here is the code and text colouring; the switcher appears once `languages` lists the translated editions.

mmcky:
Pinned theme bumped to [v2.6.1](https://github.com/QuantEcon/quantecon-theme.mystmd/releases/tag/v2.6.1), the patch release carrying the fix for QuantEcon/quantecon-theme.mystmd#186 (Back after an in-page anchor showed an "Application Error" screen on the Netlify preview; "↑ Top" did a full reload). No configuration changes. The preview built from this commit is the first real-content check of that fix.

mmcky:
Pinned theme bumped to [v2.7.0](https://github.com/QuantEcon/quantecon-theme.mystmd/releases/tag/v2.7.0) in 41b1f83c. It carries the last items of the book-theme parity gate: the pinned, scroll-tracking "On this page" panel with nested subsections (QuantEcon/quantecon-theme.mystmd#182), stderr output folded behind a "Code warnings" disclosure (QuantEcon/quantecon-theme.mystmd#92), and a layout correction for desktop windows 1280–1327px wide (QuantEcon/quantecon-theme.mystmd#198). No configuration changes are needed; this branch sets no `site.options`, so of the new OpenGraph/Twitter tags only `og:type` and `og:site_name` appear.

mmcky:
Theme **v3.0.0** is now on this branch: #629 merged into `jb2` as `69ce7e5fd`.

Three things in the description above are now out of date, listed here rather than edited in so the history stays readable.

**The toolchain table says theme `v2.2.0`.** It is `v3.0.0` as of #629 — and the table had already drifted, since `lectures/myst.yml` was pinned at v2.7.0 before this. [v3.0.0](https://github.com/QuantEcon/quantecon-theme.mystmd/releases/tag/v3.0.0) is the release that closes the book-theme parity gate.

**The `project.github` note no longer holds.** It says the theme derives the notebook-launch repo from that value plus a `.notebooks` suffix, so a stale name means broken launch buttons. v3.0.0 removed that derivation entirely: Launch is opt-in and the repository is named explicitly. `project.github` still drives the commit and "edit this page" links, so keeping it correct matters for a different reason now.

**The "Still open" bullet about launch defaults is resolved.** The URLs are no longer on theme defaults — `site.options` now carries `launch_notebook_repo`, `launch_notebook_source_dir` and `launch_colab`, ported from `lectures/_config.yml` (`nb_repository_url`, `path_to_docs`, `launch_buttons.colab_url`) rather than invented, with the branch left unset because `main` is the default. Verified on the #629 preview: the button resolves to `…/lecture-python-programming.notebooks/blob/main/need_for_speed.ipynb`, so the page location does resolve as that bullet asked someone to confirm.

Also landed in #629, for completeness: `lectures/footer.md` and the then-orphaned `_static/ccbysa.png` are gone, because v3.0.0 renders the licence footer by default and a declared footer part replaces it — the default is a superset, restoring the "A theme by QuantEcon" credit the file had dropped. `site_url` is set from the Sphinx `html.baseurl`, which is what makes the theme emit a canonical link and `og:url`. And the `{raw}` notebook-header blocks are rewritten out of `need_for_speed.md` and `polars.md`, so their source no longer reaches readers as escaped text under the page title.

mmcky:
Engine bumped in `f1c7a8d55`: the `ci.yml` fork pin moves from `96ee78f5` (2026-06-12, `qe-v8`) to `c1f274b8a` (2026-09-10, `qe-v10`). Still an exact SHA rather than a branch — the pin exists so failures stay bisectable, so only its value changed, and the comment above it now records when the SHA was taken instead of claiming it *is* the fork main.

The reason for bumping: the pin had fallen **40 commits** behind, so the HTML build was rendering with `qe-v8` while `build-ipynb.yml` — which clones `--branch main` — was already on `qe-v10`. The same lecture commit was being built by two materially different engines depending on which workflow you looked at.

Worth reading this run's log rather than just its check mark. Per the "Still open" note above, execution errors are reported without failing the build, so a green run does not by itself establish that the newer engine is clean.

mmcky:
Followed my own advice and read the log rather than the check mark. **The engine move is clean.**

Both builds pass (HTML 4m50s, ipynb 2m04s), and the bump is confirmed active in the logs: the previous run recorded `qe_version: qe-v8`, this one records `qe-v10`.

Diagnostics are unchanged across the 40 commits of engine:

| | before (`qe-v8`) | after (`qe-v10`) |
| --- | --- | --- |
| ⛔ lines | 24 | 24 |
| `error` | 6 | 6 |
| `warning` | 10 | 10 |
| unknown directive | 1 | 1 |

For the record, those 24 ⛔ lines are two known classes and neither is an execution failure: **23** are one-per-page `Resource "exports/<page>.ipynb" in downloads should be a URL or path to static file` — the `downloads` entry names an ipynb that the *HTML* build does not produce, since `build-ipynb.yml` is what creates them — and **1** is `status.md:16 unknown directive: nb-exec-table`, the Sphinx directive with no mystmd counterpart.

Worth noting against the "Still open" item above: there are currently **zero** cell-execution errors in the build, so the "errors are reported but do not fail the build" hazard is latent rather than active right now, and this green run means what it appears to mean.

PR #366, opened 2025-07-26
Adding a discussion of JAX

jstac:
Hi @mmcky , I've ported this over from the JAX GPU lecture series and tweaked it to change the focus to using JAX on CPUs.

Could you please read it carefully and make any necessary modifications -- or delegate this task? It's probably necessary to look at the text relating to timing of code execution and check that it still makes sense.

### DIFF
--- a/lectures/_config.yml
+++ b/lectures/_config.yml
@@ -15,7 +15,7 @@ latex:
      targetname: quantecon-python-programming.tex
 
 sphinx:
-  extra_extensions: [sphinx_multitoc_numbering, sphinxext.rediraffe, sphinx_tojupyter, sphinx_exercise, sphinx_togglebutton]
+  extra_extensions: [sphinx_multitoc_numbering, sphinxext.rediraffe, sphinx_tojupyter, sphinx_exercise, sphinx_togglebutton, sphinx.ext.intersphinx]
   config:
     # bibtex_reference_style: author_year #TODO: enable if bibtex bibliography is used in series
     # false-positive links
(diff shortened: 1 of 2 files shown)


mmcky:
thanks @jstac 

I will need to update the repo infrastructure to enable this series to access hardware with a `gpu`. 

Tasks

- [x] @mmcky debug missing software and target jax running on `cpu`
- [x] @mmcky read the lecture carefully in the context of this lecture series.

mmcky:
@jstac just read the `cpu` target. 

I will just have to update the software and debug. Thanks.

jstac:
Thanks Matt.  As you say, this is just for CPU.

JAX is already faster on the CPU so we can unify many of our lectures around this platform.

mmcky:
@jstac this is now running well. 

I have just completed a review. Nice, quick introduction to JAX. The difference in generating random numbers is pretty interesting and a minor inconvenience when using `jax` (having to use keys) but would assist with reproducibility with the lectures :-)

There was only one issue re: timing. There are a couple of instances where you write about removing compile time and the total run time is higher (due to higher sys time). The actual `user` time is lower but this is confusing for new programmers. 

Screenshot 2025-07-31 at 3 29 38 pm

and 
Screenshot 2025-07-31 at 3 30 20 pm

Perhaps we should remove the second part and just focus on overall execution time?

jstac:
Thanks @mmcky , nice catch.  Please go ahead with your suggestion.

PR #383, opened 2025-08-12
TEST: Timer context manager from QE.py

mmcky:
This is a test of the new Timer() context manager that is proposed for QuantEcon.py  https://github.com/QuantEcon/QuantEcon.py/pull/783

Example in `numba` lecture.


mmcky:
Closing in favour of #391

PR #437, opened 2025-11-27
ENH: Enable RunsOn GPU support for lecture builds

mmcky:
## Summary

This PR enables GPU support for building lectures using [RunsOn](https://runs-on.com/) with AWS GPU instances.

## Changes

- **Add `scripts/test-jax-install.py`** - Script to verify JAX/GPU installation works correctly
- **Add `.github/runs-on.yml`** - RunsOn configuration with QuantEcon Ubuntu 24.04 AMI (`ami-0edec81935264b6d3`) in `us-west-2`
- **Update `cache.yml`** - Use RunsOn `g4dn.2xlarge` GPU runner with JAX CUDA 13 support
- **Update `ci.yml`** - Use RunsOn `g4dn.2xlarge` GPU runner with JAX CUDA 13 support  
- **Update `publish.yml`** - Use RunsOn `g4dn.2xlarge` GPU runner with JAX CUDA 13 support

## Configuration Details

- **Runner**: `runs-on=${{ github.run_id }}/family=g4dn.2xlarge/image=quantecon_ubuntu2404/disk=large`
- **JAX**: `jax[cuda13]` with Numpyro
- **Python**: 3.13
- **AMI**: Custom QuantEcon Ubuntu 24.04 image

## Reference

This configuration mirrors the setup used in [lecture-python.myst](https://github.com/QuantEcon/lecture-python.myst).


mmcky:
## Lecture Narrative Updates

Added GPU admonitions and updated narrative text for lectures that use JAX:

| Lecture | Status | Changes Made |
|---------|--------|--------------|
| `jax_intro.md` | ✅ Updated | GPU admonition added, introduction rewritten, conditional GPU language updated |
| `numpy_vs_numba_vs_jax.md` | ✅ Updated | GPU admonition added, GPU narrative updated |
| `getting_started.md` | ✅ No change needed | Just mentions Colab/GPUs as setup options |
| `need_for_speed.md` | ✅ No change needed | Educational discussion only, no JAX code |
| `about_py.md` | ✅ No change needed | Just mentions JAX in ecosystem overview |

The GPU admonition follows the [QuantEcon style guide](https://manual.quantecon.org/styleguide/code.html#jax) and provides readers with information about GPU acceleration and alternative options for running locally.

mmcky:
@jstac I am a bit concerned by these status.html results

**GPU backend:**

Screenshot 2025-11-27 at 3 36 06 pm

---

**CPU backend (live site):**

 
Screenshot 2025-11-27 at 3 36 27 pm

mmcky:
- [x] check `status.hml` of the next build and cross check. 

https://6927d34136e0d837e87477e1--epic-agnesi-957267.netlify.app/numpy_vs_numba_vs_jax#jax-version is showing 80s for the JAX version using `lax.scan`!

mmcky:
## Hardware Benchmark Investigation Results

We ran identical benchmarks on both GitHub Actions (CPU) and RunsOn (GPU) to diagnose the performance differences.

### System Information

| Metric | GitHub Actions (CPU) | RunsOn (GPU) |
|--------|---------------------|--------------|
| **Platform** | Linux Azure | Linux AWS |
| **CPU** | AMD EPYC 7763 64-Core @ 3281 MHz | Intel Xeon Platinum 8259CL @ 2500 MHz |
| **CPU Cores** | 4 | 8 |
| **GPU** | None | Tesla T4 (15360 MiB) |

### CPU Performance (Pure Python)

| Benchmark | GitHub Actions | RunsOn | Winner |
|-----------|---------------|--------|--------|
| Integer sum (10M) | **0.599 sec** | 0.881 sec | **GitHub Actions 1.5x faster** |
| Float sqrt (1M) | **0.077 sec** | 0.107 sec | **GitHub Actions 1.4x faster** |

### CPU Performance (NumPy)

| Benchmark | GitHub Actions | RunsOn | Winner |
|-----------|---------------|--------|--------|
| Matrix multiply (3000×3000) | 0.642 sec | **0.224 sec** | **RunsOn 2.9x faster** |
| Element-wise (50M) | **1.686 sec** | 1.768 sec | ~Same |

### CPU Performance (Numba)

| Benchmark | GitHub Actions | RunsOn | Winner |
|-----------|---------------|--------|--------|
| Integer sum warm-up | 0.320 sec | **0.300 sec** | ~Same |
| Integer sum compiled | 0.000 sec | 0.000 sec | Same |
| Parallel sum warm-up | **0.344 sec** | 0.382 sec | ~Same |
| Parallel sum compiled | 0.012 sec | **0.010 sec** | ~Same |

### JAX Performance

| Benchmark | GitHub Actions (CPU) | RunsOn (GPU) | Winner |
|-----------|---------------------|--------------|--------|
| 1000×1000 warm-up | 0.030 sec | 0.079 sec | GitHub Actions |
| 1000×1000 compiled | 0.011 sec | **0.001 sec** | **GPU 11x faster** |
| 3000×3000 warm-up | 0.426 sec | 0.645 sec | GitHub Actions |
| 3000×3000 compiled | 0.276 sec | **0.009 sec** | **GPU 30x faster** |
| 50M element-wise warm-up | 0.816 sec | **0.118 sec** | **GPU 7x faster** |
| 50M element-wise compiled | 0.381 sec | **0.002 sec** | **GPU 190x faster** |

---

## Key Findings

1. **Pure Python is slower on RunsOn** - The Intel Xeon @ 2.5 GHz is slower than the AMD EPYC @ 3.3 GHz for single-threaded Python. This contributes to slower lecture execution times.

2. **NumPy matrix ops are faster on RunsOn** - Likely due to 8 cores vs 4 cores for BLAS parallelization.

3. **GPU (JAX compiled) is massively faster** - 30-190x faster for compiled operations! ✅

4. **JIT compilation overhead is higher on GPU** - Warm-up times are actually longer on GPU due to CUDA kernel compilation.

## Why `numpy_vs_numba_vs_jax` took 176 sec (GPU) vs 15 sec (CPU)

The lecture execution includes:
- **Pure Python setup code** (slower on Intel Xeon)
- **JIT compilation time** for both Numba and JAX (compilation overhead is higher for GPU kernels)
- **Multiple recompilations** when array sizes change (the lecture uses different grid sizes)
- **15 timed benchmark blocks** that each trigger fresh compilations

## Conclusion

The GPU **is working correctly** and provides massive speedups for compiled JAX operations. The slower total execution time is due to:
1. Slower single-threaded Python on the Intel Xeon CPU
2. Higher JIT compilation overhead for GPU kernels
3. The lecture code structure (many small benchmarks with varying sizes)

For production lectures that run cached notebooks, the GPU benefits will be realized. The initial build with JIT compilation will be slower, but subsequent cached runs will benefit from the pre-compiled kernels.

jstac:
This is looking good @mmcky  ---- are you still concerned?

mmcky:
@jstac yes I am concerned -- the results are strange on the `status.html` page. 

I now seeing if there is some timing issues due to Jupyter/Jupyter Kernels or Jupyter Book. 

Screenshot 2025-11-27 at 4 16 15 pm

mmcky:
## Investigation: `lax.scan` Performance on GPU

### Finding

`lax.scan` with millions of iterations performs **extremely poorly on GPU** due to a known XLA limitation. In the `numpy_vs_numba_vs_jax` lecture, the `qm_jax` function with `n=10,000,000` takes:

- **CPU**: ~0.06 seconds
- **GPU**: ~81.6 seconds (**1,340x slower**)

### Root Cause

**XLA GPU executes dynamic control flow on CPU**, requiring CPU-GPU synchronization on every loop iteration. This is documented in:

1. **JAX Issue #2491** (open since March 2020, P1 priority): [lax.scan / lax.fori_loop is slow on the GPU](https://github.com/jax-ml/jax/issues/2491)
   > *"XLA GPU currently always executes dynamic control flow on the CPU. So small loop iterations end up much slower, due to the need to frequently synchronize between the CPU/GPU."* — @shoyer (JAX collaborator)

2. **JAX Issue #29946** (July 2025): Same pattern reported with identical symptoms

### Documentation Gap

The official JAX `lax.scan` documentation does **not** warn about this GPU performance limitation. It only mentions the `unroll` parameter for partial loop unrolling, without explaining the GPU synchronization overhead.

### Impact on This PR

The GPU build is slower than CPU primarily because of this single lecture. Other JAX operations (compiled array operations, matrix multiplications) are **28-207x faster** on GPU as expected.

### Potential Solutions

1. **Reduce `n`** for GPU builds (e.g., `n=100,000` instead of `n=10,000,000`)
2. **Skip GPU execution** for this specific lecture
3. **Add documentation** in the lecture explaining this is a known XLA limitation
4. **Accept the limitation** and let the build run longer

### Recommendation

Given this is an educational resource, Option 3 (add documentation) may be most valuable—turning this limitation into a teaching moment about GPU programming patterns.

mmcky:
@jstac what is your take on this?

Here are the results of testing `cpu` and `gpu` benchmarks in both `GitHub Actions (CPU)` and `AWS EC2 RunsOn (GPU)` for the three execution pathways:
1. bare metal
2. noteboook
3. jupyter-book

they all show consistent results: https://github.com/QuantEcon/meta/issues/262#issuecomment-3584367855

I iterated with Claude and the results suggest it is to do with `lex.scan` on `gpu` along with some links to "evidence". 
https://github.com/QuantEcon/lecture-python-programming.myst/pull/437#issuecomment-3584387804. Essentially the available evidence suggest lex.scan requires to interface with `cpu` on every iteration which massively slows down computations when run on `gpu`.

@jstac I am super confused by one thing though. Your local results don't show the same when running on `gpu`?

cc: @HumphreyYang interested in your thoughts on this if you have time.

HumphreyYang:
Hi @mmcky,

I referenced this function in my unhelpful email yesterday morning which meant to discuss vectorization case and I lost focus 😬

> `lax.scan` with millions of iterations performs **extremely poorly on GPU** due to a known XLA limitation. 

I think Claude is right, and I think the problem is that

```
@partial(jax.jit, static_argnums=(1,))
def qm_jax(x0, n, α=4.0):
    def update(x, t):
        x_new = α * x * (1 - x)
        return x_new, x_new

    _, x = lax.scan(update, x0, jnp.arange(n))
    return jnp.concatenate([jnp.array([x0]), x])

```

is doing sequential scalar operation that runs on single kernel on GPU (it cannot be parallelized) so it severely underutilizes the GPU.

That's why it was comparable to `Numba` on CPU but not on GPU.

One way I often use is to keep this function off GPU to have the best of both world.

Two ways to do that are:

1. use context manager:

```
import jax
import jax.numpy as jnp
from jax import lax
from functools import partial

cpu = jax.devices("cpu")[0]

@partial(jax.jit, static_argnums=(1,))
def qm_jax(x0, n, α=4.0):
    def update(x, t):
        x_new = α * x * (1 - x)
        return x_new, x_new

    with jax.default_device(cpu):
        _, x = lax.scan(update, x0, jnp.arange(n))
        out = jnp.concatenate([jnp.array([x0]), x])
    return out
```

or 2. add it as an argument to the decorator:

```
@partial(jax.jit, static_argnums=(1,), device=cpu)
def qm_jax(x0, n, α=4.0):
    def update(x, t):
        x_new = α * x * (1 - x)
        return x_new, x_new

    _, x = lax.scan(update, x0, jnp.arange(n))
    return jnp.concatenate([jnp.array([x0]), x])
```

The result is:

image

Other parts of the lecture will still run on the GPU.

We can point out to readers why this specific piece stays on the CPU, and it actually makes a great example of one of JAX’s caveats. It might reinforces @jstac’s message about how sequential operations behave in JAX and why they often don’t benefit from GPU acceleration.

Please let me know what you think!

jstac:
Thanks @mmcky for flagging this.  Very thorough!  And super interesting.

Great analysis @HumphreyYang !!

I suggest we go for the pattern

```python
@partial(jax.jit, static_argnums=(1,), device=cpu)
def qm_jax(x0, n, α=4.0):
    def update(x, t):
        x_new = α * x * (1 - x)
        return x_new, x_new

    _, x = lax.scan(update, x0, jnp.arange(n))
    return jnp.concatenate([jnp.array([x0]), x])
```

I didn't know this pattern --- super cool.

Then let's add text around it saying why we are keeping it off the GPU!

mmcky:
Implemented the `device=cpu` pattern suggested by @HumphreyYang.

**Changes:**
- Added `cpu = jax.devices("cpu")[0]` to get the CPU device
- Modified decorator to `@partial(jax.jit, static_argnums=(1,), device=cpu)`
- Added a note explaining why we target CPU for this specific function

This keeps the `lax.scan` sequential operation on CPU where it performs well, while the rest of the lecture continues to benefit from GPU acceleration.

Commit: 54b2b34

mmcky:
thanks @jstac and @HumphreyYang. 

- [x] check run times for lecture

HumphreyYang:
Many thanks @mmcky!

I was hesitant to say that this is caused by device synchronisation because the documentation now says

"Also unlike that Python version, scan() is a JAX primitive and is lowered to a single WhileOp. "

And I think it implies that it would be a device-side only operation:

https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#conditional-graph-nodes

So I wonder if the main reason is that it is not parallelized so it runs on single thread of GPU that is much slower than a core on CPU 🤔

Maybe we should push back on Claude and see what Claude says : )

jstac:
I had the same thought @HumphreyYang

HumphreyYang:
Thanks @jstac,

I just double checked by looking at the [HLO program](https://docs.jax.dev/en/latest/aot.html). 

For this program, `main.4` is the scan code that is done in one single HLO `while` instruction:

```python
@partial(jax.jit, static_argnums=(1,))
def qm_jax(x0, n, α=4.0):
    def update(x, t):
        x_new = α * x * (1 - x)
        return x_new, x_new

    _, x = lax.scan(update, x0, jnp.arange(n))
    return jnp.concatenate([jnp.array([x0]), x])

compiled = qm_jax.lower(0.1, 5)
hlo = compiled.compiler_ir(dialect="hlo")
print(hlo.as_hlo_text())
```

```
HloModule jit_qm_jax, entry_computation_layout={(f32[])->f32[6]{0}}

closed_call.1 {
  Arg_0.1 = f32[] parameter(0)
  constant.6 = f32[] constant(4)
  mul.2 = f32[] multiply(Arg_0.1, constant.6)
  constant.5 = f32[] constant(1)
  sub.1 = f32[] subtract(constant.5, Arg_0.1)
  mul.3 = f32[] multiply(mul.2, sub.1)
  ROOT tuple.1 = (f32[], f32[]) tuple(mul.3, mul.3)
}

region_0.2 {
  arg_tuple.1 = (s32[], f32[], f32[5]{0}) parameter(0)
  get-tuple-element.3 = s32[] get-tuple-element(arg_tuple.1), index=0
  constant.7 = s32[] constant(1)
  add.1 = s32[] add(get-tuple-element.3, constant.7)
  get-tuple-element.4 = f32[] get-tuple-element(arg_tuple.1), index=1
  closed_call.3 = (f32[], f32[]) call(get-tuple-element.4), to_apply=closed_call.1
  closed_call.4 = f32[] get-tuple-element(closed_call.3), index=0
  get-tuple-element.5 = f32[5]{0} get-tuple-element(arg_tuple.1), index=2
  closed_call.5 = f32[] get-tuple-element(closed_call.3), index=1
  broadcast_in_dim.1 = f32[1]{0} reshape(closed_call.5)
  dynamic_update_slice.1 = f32[5]{0} dynamic-update-slice(get-tuple-element.5, broadcast_in_dim.1, get-tuple-element.3)
  ROOT tuple.3 = (s32[], f32[], f32[5]{0}) tuple(add.1, closed_call.4, dynamic_update_slice.1)
}

region_1.3 {
  arg_tuple.3 = (s32[], f32[], f32[5]{0}) parameter(0)
  get-tuple-element.10 = f32[] get-tuple-element(arg_tuple.3), index=1
  get-tuple-element.11 = f32[5]{0} get-tuple-element(arg_tuple.3), index=2
  get-tuple-element.9 = s32[] get-tuple-element(arg_tuple.3), index=0
  constant.9 = s32[] constant(5)
  ROOT lt.1 = pred[] compare(get-tuple-element.9, constant.9), direction=LT
}

ENTRY main.4 {
  constant.10 = s32[] constant(0)
  x0.1 = f32[] parameter(0)
  constant.11 = f32[] constant(0)
  broadcast.1 = f32[5]{0} broadcast(constant.11), dimensions={}
  while.5 = (s32[], f32[], f32[5]{0}) tuple(constant.10, x0.1, broadcast.1)
  while.6 = (s32[], f32[], f32[5]{0}) while(while.5), condition=region_1.3, body=region_0.2
  while.7 = s32[] get-tuple-element(while.6), index=0
  while.8 = f32[] get-tuple-element(while.6), index=1
  broadcast_in_dim.3 = f32[1]{0} reshape(x0.1)
  while.9 = f32[5]{0} get-tuple-element(while.6), index=2
  ROOT concatenate.1 = f32[6]{0} concatenate(broadcast_in_dim.3, while.9), dimensions={0}
}
```

So I am 95% sure it is a device-side operation.

I will update the nice note @mmcky wrote!

HumphreyYang:
Hi @jstac and @mmcky, I just pushed in an updated note. Please feel free to edit or let me know how to improve my note!

mmcky:
thanks @HumphreyYang nice change. I like it. 

One last comment, for the run time to go from `0.06s` to `83s` there has to be a hardware issue going on as well (like oscillating between `cpu` and `gpu`) but I agree not being 100% certain this is a much better `note`.

Nice work @HumphreyYang

mmcky:
Thanks @HumphreyYang for the HLO analysis! 

**Regarding the HLO output:**

The HLO shows `lax.scan` compiles to a single `while` instruction (`while.6` in `main.4`), which represents the *logical* computation structure. However, the HLO is device-agnostic - it shows what computation to do, not where/how it executes at runtime.

**The mystery remains:**

If the while loop is truly a single device-side operation, we wouldn't expect the 81s GPU vs 0.06s CPU difference. The JAX Issue #2491 suggests XLA's GPU implementation of while loops may require the loop condition to be evaluated on CPU, causing sync on each iteration - but this wouldn't be visible in the HLO.

**Possible next steps to investigate:**

1. **NVIDIA Nsight Systems profiling** - Would show actual CPU-GPU transfers:
   ```bash
   nsys profile -o scan_profile python -c "..."
   ```

2. **JAX profiler** - Can trace device placement:
   ```python
   with jax.profiler.trace("/tmp/jax-trace"):
       result = qm_jax(0.1, n).block_until_ready()
   ```

3. **Check XLA debug logs**:
   ```python
   import os
   os.environ["XLA_FLAGS"] = "--xla_dump_to=/tmp/xla_dump"
   ```

The `device=cpu` fix works empirically, but understanding the root cause could be valuable for documentation and for cases where forcing CPU isn't desirable.

mmcky:
**Next TODO:**

- [x] run profiler to check `lex.scan` to further identify root cause (hardware side of things). 

**Final TODO:**

- [x] @jstac to review `content` suggestions.

HumphreyYang:
Many thanks @mmcky, that's great to know. I always thought each HLO block will run on the same device. 

I am looking forward to seeing the result for your benchmark!

mmcky:
@HumphreyYang here is the initial diagnostic output. It's replicating the condition, but not sure I fully agree with the diagnostic as definitive proof. This is now just for fun. I like your new note. 👍 

```
============================================================
lax.scan GPU Performance Profiling
============================================================
JAX version: 0.8.1
Available devices: [CudaDevice(id=0)]
Default device: cuda:0
Iterations (n): 100,000
--- Compilation (warm-up) ---
Compiling default device version... done (1.12s)
Compiling CPU version... done (0.06s)
--- Timing Runs (post-compilation) ---
Default device (cuda:0):
  Run 1: 0.804583s
  Run 2: 0.805552s
  Run 3: 0.804225s
CPU (forced with device=cpu):
  Run 1: 0.000848s
  Run 2: 0.000767s
  Run 3: 0.000541s
============================================================
SUMMARY
============================================================
Default device avg: 0.804787s
CPU avg:            0.000718s
Ratio (default/cpu): 1120.4x
⚠️  GPU is significantly slower than CPU!
   This confirms the lax.scan synchronization overhead issue.
============================================================
DIAGNOSTIC: Per-iteration Sync Overhead Analysis
============================================================
If there's a CPU-GPU sync per iteration, time should scale
linearly with iteration count (not with compute work).
Iteration Count | GPU Time (s) | Time/Iter (µs) | Expected if O(n)
----------------------------------------------------------------------
          1,000 |     0.008302 |           8.30 | 0.008302
          5,000 |     0.040774 |           8.15 | 0.041511
         10,000 |     0.080552 |           8.06 | 0.083021
         50,000 |     0.401472 |           8.03 | 0.415105
        100,000 |     0.805949 |           8.06 | 0.830211
Scaling analysis:
  Time ratio (100k/1k iterations): 97.1x
  Expected if linear O(n): 100.0x
✓ Time scales ~linearly with iterations!
  This indicates constant per-iteration overhead (CPU-GPU sync).
  Estimated sync overhead: ~8.3 µs per iteration
The diagnostic shows if time scales linearly with iterations,
which indicates constant per-iteration CPU-GPU sync overhead.
```

I have added nsight --profiler output to be uploaded as an artifact. However I just realised I won't be able to see the results file as I don't have a computer with `nsight ui` tool loaded. Doh! :-). I'll let you know where it is if you want to take a look.

HumphreyYang:
Thanks @mmcky for the potential next steps! (I think I clicked send at the same time as you post the result!)

I followed your suggestion to use tensorboard. Here is the result:

image


I think it shows that the operations are mainly done on the device!

HumphreyYang:
Tensorboard is cool:

image

Looks like the host is in charge of compiling and it takes a lot of time:

| Rank | Host/Device | Operation Type       | Operation Name                              | #Occurrences | Total time (µs) | Avg. time (µs) | Total self-time (µs) | Avg. self-time (µs) | Device Self-Time (%) | Device Cumulative (%) | Host Self-Time (%) | Host Cumulative (%) | Normalized FLOP Rate | Model FLOP Rate (GFLOP/s) | Memory BW (GB/s) | Op Intensity (FLOPs/Byte) | TensorCore Utilization | Bound by | Exec Mode |
| ---: | ----------- | -------------------- | ------------------------------------------- | -----------: | --------------: | -------------: | -------------------: | ------------------: | -------------------: | --------------------: | -----------------: | ------------------: | -------------------: | ------------------------: | ---------------: | ------------------------: | ---------------------: | -------- | --------- |
|    1 | Device      | dynamic_update_slice | jit(qm_jax)/while/body/dynamic_update_slice |        1,000 |             851 |              1 |                  851 |                   1 |                33.7% |                 33.7% |                 0% |                  0% |                    0 |                         0 |            0.013 |                         0 |                     0% | HBM      | Function  |
|    2 | Device      | mul                  | jit(qm_jax)/while/body/closed_call/mul      |        1,000 |             845 |              1 |                  845 |                   1 |                33.4% |                 67.1% |                 0% |                  0% |                0.011 |                     0.011 |            0.009 |                     1.125 |                     0% | HBM      | Function  |
|    3 | Device      | add                  | jit(qm_jax)/while/body/add                  |        1,000 |             827 |              1 |                  827 |                   1 |                32.7% |                 99.9% |                 0% |                  0% |                0.004 |                     0.004 |            0.009 |                     0.375 |                     0% | HBM      | Function  |
|    4 | Device      | Unknown              | copy.11                                     |            1 |               1 |              1 |                    1 |                   1 |                   0% |                 99.9% |                 0% |                  0% |                    0 |                         0 |            0.007 |                         0 |                     0% | HBM      | Function  |
|    5 | Device      | Unknown              | copy.2                                      |            1 |               1 |              1 |                    1 |                   1 |                   0% |                 99.9% |                 0% |                  0% |                    0 |                         0 |            0.008 |                         0 |                     0% | HBM      | Function  |
|    6 | Device      | concatenate          | jit(qm_jax)/concatenate                     |            1 |               1 |              1 |                    1 |                   1 |                   0% |                  100% |                 0% |                  0% |                6.256 |                     6.256 |            7.769 |                      0.75 |                     0% | HBM      | Function  |
|    7 | Device      | broadcast_in_dim     | jit(qm_jax)/broadcast_in_dim                |            1 |               1 |              1 |                    1 |                   1 |                   0% |                  100% |                 0% |                  0% |                    0 |                         0 |            5.292 |                         0 |                     0% | HBM      | Function  |
|    8 | Host        | jit_qm_jax           | Compiling IR                                |            1 |          17,282 |         17,282 |               17,282 |              17,282 |                   0% |                  100% |              66.2% |               66.2% |                    0 |                         0 |                0 |                         0 |                     0% | Unknown  | Function  |
|    9 | Host        | jit_qm_jax           | HLO Transforms                              |            1 |           8,841 |          8,841 |                8,841 |               8,841 |                   0% |                  100% |              33.8% |                100% |                    0 |                         0 |                0 |                         0 |                     0% | Unknown  | Function  |

mmcky:
## Nsight Systems Artifact Available 🎉

@HumphreyYang the Nsight profile artifact is now available for download. Would love your help analyzing it!

### How to Use the Artifact

1. **Download**: Go to Actions → latest workflow run → Artifacts section → download `nsight-profile.zip`
2. **Extract**: Get `lax_scan_trace.nsys-rep`
3. **Open**: Use Nsight Systems UI (`nsys-ui lax_scan_trace.nsys-rep`)

### What to Look For

The key question: **Is the ~8µs per iteration from CPU-GPU sync, kernel launch overhead, or single-threaded GPU execution?**

**Timeline patterns to identify:**

| Pattern | Interpretation |
|---------|----------------|
| Long continuous GPU bar | Loop runs entirely on GPU (single-threaded GPU issue) |
| Many tiny GPU bars with gaps | Kernel launch overhead per iteration |
| CPU activity interleaved with GPU | CPU-GPU synchronization per iteration |
| Repeated `cudaMemcpy` calls | Data transfers between host and device |
| `cudaDeviceSynchronize` calls | Explicit sync points |

**Expected if CPU-GPU sync per iteration:**
```
CPU:  ████░░████░░████░░████░░...  (busy-wait-busy-wait pattern)
GPU:     █     █     █     █  ...  (tiny kernels with gaps)
```

**Expected if single-threaded GPU (no sync):**
```
CPU:  █░░░░░░░░░░░░░░░░░░░░░░░█    (idle during execution)
GPU:  ██████████████████████████   (one long continuous kernel)
```

### Key Metrics to Note
- Total CUDA API calls count
- Average kernel duration  
- GPU utilization percentage
- Any `cudaStreamSynchronize` or `cudaDeviceSynchronize` counts

This should definitively show whether the ~8µs per iteration overhead is from CPU-GPU sync, kernel launch overhead, or something else!

lectures/jax_intro.md
```diff
@@ -21,20 +21,23 @@ In addition to what's in Anaconda, this lecture will need the following librarie
...
-However, JAX seamlessly handles transitions across different hardware platforms.
+This lecture provides a short introduction to [Google JAX](https://github.com/jax-ml/jax).
 
-As a result, if you run this code on a machine with a GPU and a GPU-aware
-version of JAX installed, your code will be automatically accelerated and you
-will receive the full benefits.
+JAX provides a NumPy-like interface that can leverage GPU acceleration for high-performance numerical computing.
 
-For a discussion of JAX on GPUs, see [our JAX lecture series](https://jax.quantecon.org/intro.html).
+For a more comprehensive discussion of JAX, see [our JAX lecture series](https://jax.quantecon.org/intro.html).
```
jstac: I suggest deleting this line --- we'll remove that whole site before too long.

mmcky:
## Analysis of @HumphreyYang's TensorBoard Results 🔍

Great analysis @HumphreyYang! The TensorBoard data is very revealing.

### Key Findings from the Operation Statistics

| Operation | Location | Count | Time (µs) | % of Device Time |
|-----------|----------|-------|-----------|------------------|
| `dynamic_update_slice` | Device | 1,000 | 851 | 33.7% |
| `mul` | Device | 1,000 | 845 | 33.4% |
| `add` | Device | 1,000 | 827 | 32.7% |
| `Compiling IR` | **Host** | 1 | 17,282 | (compilation) |
| `HLO Transforms` | **Host** | 1 | 8,841 | (compilation) |

### What This Tells Us

1. **Operations ARE on Device** ✅ - The `mul`, `add`, and `dynamic_update_slice` all run on GPU with 1,000 calls each (matching iteration count)

2. **Each iteration launches 3 separate kernels**:
   - `dynamic_update_slice` (writing result to array)
   - `mul` (the α * x * (1-x) computation)  
   - `add` (loop counter increment)

3. **~0.85µs per kernel average** - Each operation takes roughly 0.85µs

4. **Math check**: 3 kernels × ~0.85µs × 1000 = ~2.5ms device time, but we measure ~8ms total

### Root Cause: Kernel Launch Overhead

The gap between device time (~2.5ms) and wall time (~8ms) is **kernel launch overhead**!

- GPU kernel launch overhead is typically ~2-10µs per launch
- With 3 kernels per iteration: 3 × ~2-3µs overhead = ~6-9µs per iteration
- This matches our measured **~8µs per iteration**!

### Why This Happens

The HLO `while` loop doesn't fuse into a single kernel. XLA generates **separate kernels for each operation inside the loop body**, and each kernel launch has fixed overhead regardless of how tiny the actual compute is.

### Conclusion

- ❌ **Not CPU-GPU data synchronization** per iteration
- ✅ **Kernel launch overhead** from launching 3 tiny kernels per iteration
- The compute itself is fast (~0.85µs), but launching 3 kernels adds ~6µs overhead per iteration

This is actually a cleaner explanation! The `device=cpu` fix avoids the kernel launch overhead entirely by running the sequential loop on CPU where there's no kernel launch cost.

Thanks for the great detective work! 🕵️

jstac:
This is great guys.  Fascinating discussion.  It's fun working with such a good team :-)

One small comment but other than that the changes are really nice.

jstac:
> The HLO while loop doesn't fuse into a single kernel. XLA generates separate kernels for each operation inside the loop body, and each kernel launch has fixed overhead regardless of how tiny the actual compute is.

:scream:

HumphreyYang:
> > The HLO while loop doesn't fuse into a single kernel. XLA generates separate kernels for each operation inside the loop body, and each kernel launch has fixed overhead regardless of how tiny the actual compute is.
> 
> 😱

Yeah this sounds really bad : ( 

We have to have large computation each loop to worth the overhead.

lectures/jax_intro.md
mmcky: thanks @jstac done.

HumphreyYang:
Hi @mmcky, 

For your interest, here is the output from `nsight`!

image

mmcky:
## Nsight Systems Confirmation

Thanks @HumphreyYang for running the Nsight visualization! 🎉

The timeline view perfectly confirms our kernel launch overhead hypothesis:

- Each `lax.scan` iteration launches **3 separate GPU kernels** (mul, add, dynamic_update_slice)
- The visualization shows the characteristic pattern of many tiny kernel launches with gaps between them
- Those gaps represent the kernel launch latency (~2-3µs per launch)
- With 3 kernels per iteration × ~2-3µs each = **~6-9µs overhead per iteration**

This matches our measured ~8µs per iteration and explains the 1000x+ slowdown for scalar operations.

**Key takeaway for the lectures:** Sequential scalar operations in `lax.scan` should use `device=cpu` to avoid this overhead. GPU excels when each kernel does substantial parallel work, not when launching millions of tiny kernels.

mmcky:
thanks @jstac and @HumphreyYang -- learnt lots today. Been fun. Appreciate iterating with you both.

mmcky:
@jstac In the `numpy_vs_numba_vs_jax` lecture, the statement about vmap version 2:

> "We don't get much speed gain but we do save some memory."

This appears to be **incorrect** based on the GPU build results:

- **vmap Version 1**: 0.00111055 seconds (1.1 ms)
- **vmap Version 2**: 0.00046182 seconds (0.46 ms)

Version 2 is actually **~2.4x faster** than Version 1, in addition to being more memory efficient.

This makes sense—by taking the `max` inside the inner function, we avoid creating large intermediate arrays and the GPU can compute row maxes immediately.

- [x] update to reflect that Version 2 is both faster AND more memory efficient.

Is this a different result to what we had on `cpu`?

mmcky:
> Is this a different result to what we had on `cpu`? 

Ah yes this is different between `cpu` and `gpu`. When run on `cpu` they have the same times. 

https://python-programming.quantecon.org/numpy_vs_numba_vs_jax.html#vmap-version-2

mmcky:
**Final TODO:**

- [x] clean up `scripts` and debug infrastructure from `ci.yml`
- [x] minor update to `vmap version 2` narrative.

mmcky:
## Timing Comparison Review - `numpy_vs_numba_vs_jax` Lecture

I reviewed all timing comparisons in the lecture. Here's the summary:

### ✅ Statements that are correct:
1. **NumPy vs Numba (non-parallel)**: "can be a bit slower or a bit faster" - appropriately hedged
2. **Numba parallel**: "should see at least some benefits" - correct (4.7x speedup)
3. **JAX meshgrid**: "significantly faster than NumPy" - correct (13x faster)
4. **Sequential operations**: "Both JAX and Numba deliver strong performance" - correct (0.065s vs 0.069s)

### ⚠️ Potentially questionable:
**vmap Version 1** states: "The execution time is similar to the mesh operation"
- Meshgrid JAX: 0.01924109 seconds
- vmap Version 1: 0.00111055 seconds
- vmap v1 is actually **~17x faster** than meshgrid, not "similar"

### ❌ Incorrect (already reported):
**vmap Version 2** states: "We don't get much speed gain but we do save some memory"
- vmap v1: 0.00111055 seconds  
- vmap v2: 0.00046182 seconds
- vmap v2 is actually **~2.4x faster**, not "not much speed gain"

---

**Recommendation**: Update both vmap statements to reflect the actual performance improvements shown in the GPU build results.

jstac:
## Pedagogical improvements to JAX lectures

This commit improves the structure and clarity of the JAX introduction and vmap sections:

**jax_intro.md:**
- Reorganized to introduce JAX's key features (parallelization, JIT compilation, autodiff) upfront before installation
- Improved flow by presenting "what is JAX" before "how to install JAX"
- Refined language around JAX as a NumPy drop-in replacement

**numpy_vs_numba_vs_jax.md:**
- Added explicit GPU performance notes for vmap implementations
- Enhanced explanation of vmap version 2 with detailed breakdown of function composition (`f_vec_x_max` → `f_vec_max` → `jnp.max`)
- Clarified memory efficiency tradeoffs between different vmap approaches
- Updated performance expectations to reflect GPU execution context

These changes make the lectures more accessible while better highlighting the GPU acceleration benefits that this PR enables.

lectures/numpy_vs_numba_vs_jax.md
```diff
@@ -399,23 +417,28 @@ def compute_max_vmap_v2(grid):
...
 Let's run it again to eliminate compilation time:
 
 ```{code-cell} ipython3
 with qe.Timer(precision=8):
     z_max = compute_max_vmap_v2(grid).block_until_ready()
 ```
 
-We don't get much speed gain but we do save some memory.
-
+If you are running this on a GPU, as well are, you should see another nontrivial speed gain.
```
mmcky:
```suggestion
If you are running this on a GPU, as we are, you should see another nontrivial speed gain.
```

mmcky: @jstac just a minor typo.

mmcky:
Thanks @jstac for your edits. This is ready to go. 

I will:
1. merge
2. manually run the cache to get the latest execution cache in the correct environment
3. run a publish

PR #442, opened 2025-11-27
FIX: jax_intro timeout: use lax.fori_loop instead of Python for loop

### LINKED ISSUE #441: BUG: Cell timeout in jax_intro lecture
The `cache` is reporting execution failures on `jax_intro` 

[jax_intro.err.log](https://github.com/user-attachments/files/23811456/jax_intro.err.log)

```
with qe.Timer():
    compute_call_price_jax().block_until_ready()
```

- [x] trigger new run
- [x] investigate further if required

mmcky:
## Problem

The `compute_call_price_jax` function in `jax_intro.md` was timing out during `cache.yml` builds (600s cell execution timeout).

fixes #441 

## Root Cause

JAX unrolls Python `for` loops during JIT compilation. With large arrays (`M=10,000,000`), this causes excessive compilation time as JAX traces through each iteration separately.

## Solution

Replace the Python `for` loop with `jax.lax.fori_loop`, which compiles the loop efficiently without unrolling:

```python
# Before (Python for loop - gets unrolled)
for t in range(n):
    key, subkey = jax.random.split(key)
    Z = jax.random.normal(subkey, (2, M))
    s = s + μ + jnp.exp(h) * Z[0, :]
    h = ρ * h + ν * Z[1, :]

# After (JAX fori_loop - compiled efficiently)
def loop_body(i, state):
    s, h, key = state
    key, subkey = jax.random.split(key)
    Z = jax.random.normal(subkey, (2, M))
    s = s + μ + jnp.exp(h) * Z[0, :]
    h = ρ * h + ν * Z[1, :]
    return s, h, key

s, h, key = jax.lax.fori_loop(0, n, loop_body, (s, h, key))
```

Added an explanatory note for students about why we use `fori_loop`.

## Related

This is the same category of issue as the `lax.scan` GPU performance problem in PR #437, but with a different manifestation (compilation time vs runtime).

### DIFF
--- a/lectures/jax_intro.md
+++ b/lectures/jax_intro.md
@@ -833,15 +833,18 @@ def compute_call_price_jax(β=β,
     s = jnp.full(M, np.log(S0))
     h = jnp.full(M, h0)
 
-    def loop_body(i, state):
-        s, h, key = state
+    def update(i, loop_state):
+        s, h, key = loop_state
         key, subkey = jax.random.split(key)
         Z = jax.random.normal(subkey, (2, M))
         s = s + μ + jnp.exp(h) * Z[0, :]
         h = ρ * h + ν * Z[1, :]
-        return s, h, key
+        new_loop_state = s, h, key
+        return new_loop_state
 
-    s, h, key = jax.lax.fori_loop(0, n, loop_body, (s, h, key))
+    loop_state = s, h, key
+    final_loop_state = jax.lax.fori_loop(0, n, update, loop_state)
+    s, h, key = final_loop_state
 
     expectation = jnp.mean(jnp.maximum(jnp.exp(s) - K, 0))
 
--- a/lectures/jax_intro.md
+++ b/lectures/jax_intro.md
@@ -832,14 +832,26 @@ def compute_call_price_jax(β=β,
 
     s = jnp.full(M, np.log(S0))
     h = jnp.full(M, h0)
-    for t in range(n):
+
+    def loop_body(i, state):
+        s, h, key = state
         key, subkey = jax.random.split(key)
         Z = jax.random.normal(subkey, (2, M))
         s = s + μ + jnp.exp(h) * Z[0, :]
         h = ρ * h + ν * Z[1, :]
+        return s, h, key
+
+    s, h, key = jax.lax.fori_loop(0, n, loop_body, (s, h, key))
+
     expectation = jnp.mean(jnp.maximum(jnp.exp(s) - K, 0))
 
     return β**n * expectation
+```
+
+```{note}
+We use `jax.lax.fori_loop` instead of a Python `for` loop.
+This allows JAX to compile the loop efficiently without unrolling it,
+which significantly reduces compilation time for large arrays.
 ```
 
 Let's run it once to compile it:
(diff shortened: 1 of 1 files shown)


jstac:
Thanks @mmcky !  Useful catch!

As a standard convention, I've been using the following

 `loop_body`  -> `update`
`state` -> `loop_state`
I might also add `new_loop_state = ...` before `return new_loop_state` and also `final_loop_state = ...` followed by unpacking.

It's a bit verbose but this is the first time readers will see a fori_loop.

mmcky:
## Timing Comparison

### PR #442 (`fori_loop` on GPU)
- **First run (compile + execute):** 1.86 seconds
- **Second run (cached):** 0.48 seconds

### Production (Python `for` loop on CPU)
- **First run (compile + execute):** 16.73 seconds
- **Second run (cached):** 14.37 seconds

### Summary

| Version | First Run | Second Run | Improvement |
|---------|-----------|------------|-------------|
| Old (`for` loop, CPU) | 16.73s | 14.37s | baseline |
| New (`fori_loop`, GPU) | 1.86s | 0.48s | **~9x faster compile, ~30x faster run** |

The `fori_loop` fix works well:

1. **Compilation time reduced from ~17s to ~2s** - JAX no longer unrolls 20 iterations with M=10,000,000 arrays
2. **Runtime reduced from ~14s to ~0.5s** - GPU acceleration now works effectively
3. **No more timeout** - the old version was timing out at 600s in `cache.yml` builds; this completes in under 2.5s total

mmcky:
Thanks @jstac this overhead thing in `jax` makes a pretty big difference.

lectures/jax_intro.md
```diff
@@ -833,15 +833,18 @@ def compute_call_price_jax(β=β,
...
         key, subkey = jax.random.split(key)
         Z = jax.random.normal(subkey, (2, M))
         s = s + μ + jnp.exp(h) * Z[0, :]
         h = ρ * h + ν * Z[1, :]
-        return s, h, key
+        new_loop_state = s, h, key
+        return new_loop_state
 
-    s, h, key = jax.lax.fori_loop(0, n, loop_body, (s, h, key))
+    loop_state = s, h, key
```
jstac: i missed one: `loop_state` on 845 should be `initial_loop_state` according to convention.

sorry @mmcky :grimacing:

jstac:
> Thanks @jstac this overhead thing in `jax` makes a pretty big difference.

Thanks for all this analysis @mmcky .  Super useful.  JAX unrolls a regular for loop, leading to massive compile times.  So glad you picked this up!

(Some things were written in the early times, when we just got started with JAX.)

jstac:
I'll stay out of this now, many thanks @mmcky !

mmcky:
thanks @jstac. Done [9bceb2d](https://github.com/QuantEcon/lecture-python-programming.myst/pull/442/commits/9bceb2d135befba14faf2c11ad95aae9009f4510). 

Appreciate the review. 

---

This process has led me to open this project idea: https://github.com/QuantEcon/meta/issues/264. When using `dask` I always appreciated the `group execution into chunks > 10ms` guideline to improve efficiency. I think a detailed set of benchmarks would be really neat to document overheads and how to use packages most effectively - that get's updated as versions are released.

PR #525, opened 2026-04-12
Improve NumPy vs Numba vs JAX lecture

jstac:
## Summary
- Rewrite prange section: explain why the broken version always returns `-inf` (unrecognized reduction pattern, not a classical race condition), add `max()` fix, keep `row_maxes` alternative
- Restructure vmap into a single narrative: memory savings first, then kernel fusion via wrapping in `@jax.jit`
- Remove unsupported speed claims (Numba vs JAX for sequential ops, vmap "significantly faster on GPU")
- Move `lax.scan` autodiff advantage into a `{note}`
- Fix `(parallel)=` label → `(numpy_numba_jax)=`, `Matlab` → `MATLAB`

## Test plan
- [x] Converted to .py via jupytext and ran all cells — executes cleanly
- [ ] Build the lecture site and verify page renders correctly
- [ ] Verify `{note}` directive displays properly

### DIFF
--- a/lectures/numpy_vs_numba_vs_jax.md
+++ b/lectures/numpy_vs_numba_vs_jax.md
@@ -9,7 +9,7 @@ kernelspec:
   name: python3
 ---
 
-(parallel)=
+(numpy_numba_jax)=
 ```{raw} jupyter
 <div id="qe-notebook-header" align="right" style="text-align:right;">
         <a href="https://quantecon.org/" title="quantecon.org">
@@ -141,7 +141,7 @@ code executes relatively quickly.
 Here we use `np.meshgrid` to create two-dimensional input grids `x` and `y` such
 that `f(x, y)` generates all evaluations on the product grid.
 
-(This strategy dates back to Matlab.)
+(This strategy dates back to MATLAB.)
 
 ```{code-cell} ipython3
 grid = np.linspace(-3, 3, 3_000)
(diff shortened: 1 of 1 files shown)


jstac:
## What we changed and why

**prange reduction explanation rewritten:** The original text described the broken `prange` example as a classical race condition ("threads read stale values and overwrite each other's updates"). But the code always returns `-inf`, which points to a different mechanism: Numba's `prange` creates thread-private copies of `m`, each correctly updated within its chunk, but the `if z > m: m = z` pattern isn't recognized as a max reduction. Without a recognized combining operator, the per-thread results are never merged and `m` retains its initial value. The rewrite explains this accurately and adds the simple `max()` fix (which Numba does recognize as a reduction), keeping the `row_maxes` approach as an alternative that handles the reduction explicitly.

**vmap sections restructured:** Previously split into "Version 1" and "Version 2" subsections. Now a single narrative: vmap avoids large input arrays (memory savings, similar runtime) → but we're leaving speed on the table because `jnp.max` sits outside the JIT → fix by pushing max inside and wrapping in a single `@jax.jit` (kernel fusion, even more memory savings). This makes the progression logical rather than presenting two versions without explaining why we need the second.

**Unsupported speed claims removed:**
- "vmap is usually significantly faster on GPU" — actually slightly slower in practice
- "Numba typically offering slightly better speeds on purely sequential operations" — JAX is comparable or slightly faster
- "Numba is the clear winner... as well as high performance" — trimmed to focus on code clarity, which is the real advantage
- "performance gain is minimal for purely sequential work" — removed the implicit claim that Numba is faster

**`lax.scan` autodiff:** Moved into a `{note}` — it's an important aside but shouldn't interrupt the main comparison flow.

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python-programming.zh-cn/pull/49
**Files synced** (1):
- `lectures/numpy_vs_numba_vs_jax.md`

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python-programming.fa/pull/96
**Files synced** (1):
- `lectures/numpy_vs_numba_vs_jax.md`

PR #550, opened 2026-05-11
[numba.md] Update np.random → Generator API

Chihiro2000GitHub:
## Summary

This PR migrates legacy NumPy random API usage in `numba.md` as part of [QuantEcon/meta#299](https://github.com/QuantEcon/meta/issues/299).

## Details

The Numba/JIT-related classification and changes in this PR follow the guidance at https://manual.quantecon.org/styleguide/code.html#numpy-random-number-generation. I would be grateful if you could refer to the Numba section of this page when reviewing.

**Case A** (`update()`, main text, `@jit`): Left unchanged. Although `update()` is `@jit`-decorated, it is called from a `prange` loop in `compute_long_run_median_parallel`, making it unsafe to pass a shared Generator. Flagged for reviewer judgment.

**Case B** (speed_ex1 solution, `@jit` / no parallel): `rng = np.random.default_rng()` placed before the `@jit` definition; signature changed to `calculate_pi(rng, n=...)`; `np.random.uniform` → `rng.uniform`. Call sites updated. Note: passing a Generator into a `@jit` function may require Numba to use object mode. I checked that the updated code runs as expected on my side, but I would appreciate reviewer confirmation.

**Case C** (speed_ex2 solution, plain Python / later JIT-compiled via `jit(compute_series)`): Same pattern as Case B. Signature changed to `compute_series(n, rng)`; `np.random.uniform` → `rng.uniform`. Both the pure Python and `jit`-compiled call sites updated.

**Case D** (numba_ex3 solution, `@jit(parallel=True)` + `prange`): Draws lifted outside the `prange` loop. `rng`, `u_draws`, and `v_draws` defined before the `@jit(parallel=True)` definition; signature changed to `calculate_pi(u_draws, v_draws)`; loop length derived from `len(u_draws)`. `n = 1_000_000` kept to match the original example size.

**Case E** (numba_ex4 solution, `@jit(parallel=True)` + `prange`): Legacy `np.random.randn()` calls kept intentionally as a narrow memory-constrained exception. Pre-allocating `(M, n)` shock arrays with `M = 10_000_000` and `n = 20` would require approximately 3.2 GB, which is likely beyond what most readers' machines can comfortably accommodate. A short comment has been added inside the loop explaining this.

Hi @mmcky and @HumphreyYang, I'd be grateful if you could take a look when you have time.

### DIFF
--- a/lectures/numba.md
+++ b/lectures/numba.md
@@ -454,11 +454,13 @@ Compare speed with and without Numba when the sample size is large.
 Here is one solution:
 
 ```{code-cell} ipython3
+rng = np.random.default_rng()
+
 @jit
-def calculate_pi(n=1_000_000):
+def calculate_pi(rng, n=1_000_000):
     count = 0
     for i in range(n):
-        u, v = np.random.uniform(0, 1), np.random.uniform(0, 1)
+        u, v = rng.uniform(0, 1), rng.uniform(0, 1)
         d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
         if d < 0.5:
             count += 1
@@ -471,12 +473,12 @@ Now let's see how fast it runs:
 
 ```{code-cell} ipython3
 with qe.Timer():
-    calculate_pi()
+    calculate_pi(rng)
 ```
 
 ```{code-cell} ipython3
 with qe.Timer():
-    calculate_pi()
+    calculate_pi(rng)
 ```
 
 If we switch off JIT compilation by removing `@jit`, the code takes around
@@ -550,10 +552,12 @@ p, q = 0.1, 0.2  # Prob of leaving low and high state respectively
 Here's a pure Python version of the function
 
 ```{code-cell} ipython3
-def compute_series(n):
+rng = np.random.default_rng()
+
+def compute_series(n, rng):
     x = np.empty(n, dtype=np.int64)
     x[0] = 1  # Start in state 1
-    U = np.random.uniform(0, 1, size=n)
+    U = rng.uniform(0, 1, size=n)
     for t in range(1, n):
         current_x = x[t-1]
         if current_x == 0:
@@ -568,7 +572,7 @@ state is about 0.666
 
 ```{code-cell} ipython3
 n = 1_000_000
-x = compute_series(n)
+x = compute_series(n, rng)
 print(np.mean(x == 0))  # Fraction of time x is in state 0
 ```
 
@@ -578,7 +582,7 @@ Now let's time it:
 
 ```{code-cell} ipython3
 with qe.Timer():
-    compute_series(n)
+    compute_series(n, rng)
 ```
 
 Next let's implement a Numba version, which is easy
@@ -590,7 +594,7 @@ compute_series_numba = jit(compute_series)
 Let's check we still get the right numbers
 
 ```{code-cell} ipython3
-x = compute_series_numba(n)
+x = compute_series_numba(n, rng)
 print(np.mean(x == 0))
 ```
 
@@ -598,7 +602,7 @@ Let's see the time
 
 ```{code-cell} ipython3
 with qe.Timer():
-    compute_series_numba(n)
+    compute_series_numba(n, rng)
 ```
 
 This is a nice speed improvement for one line of code!
@@ -636,11 +640,17 @@ For the size of the Monte Carlo simulation, use something substantial, such as
 Here is one solution:
 
 ```{code-cell} ipython3
+n = 1_000_000
+rng = np.random.default_rng()
+u_draws = rng.uniform(size=n)
+v_draws = rng.uniform(size=n)
+
 @jit(parallel=True)
-def calculate_pi(n=1_000_000):
+def calculate_pi(u_draws, v_draws):
+    n = len(u_draws)
     count = 0
     for i in prange(n):
-        u, v = np.random.uniform(0, 1), np.random.uniform(0, 1)
+        u, v = u_draws[i], v_draws[i]
         d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
         if d < 0.5:
             count += 1
@@ -653,12 +663,12 @@ Now let's see how fast it runs:
 
 ```{code-cell} ipython3
 with qe.Timer():
-    calculate_pi()
+    calculate_pi(u_draws, v_draws)
 ```
 
 ```{code-cell} ipython3
 with qe.Timer():
-    calculate_pi()
+    calculate_pi(u_draws, v_draws)
 ```
 
 By switching parallelization on and off (selecting `True` or
@@ -773,6 +783,7 @@ def compute_call_price_parallel(β=β,
         s = np.log(S0)
         h = h0
         # Simulate forward in time
+        # Draws are kept inside the loop to avoid pre-allocating large shock arrays.
         for t in range(n):
             s = s + μ + np.exp(h) * np.random.randn()
             h = ρ * h + ν * np.random.randn()
--- a/lectures/numba.md
+++ b/lectures/numba.md
@@ -863,7 +863,7 @@ ax.legend()
 plt.show()
 ```
 
-Both bands are centered on $\pi$, but the band associated with the data race is much wider than the other one and narrows very slowly as the sample size grows.
+Both bands are centered on $\pi$, but the band associated with the data race is wider than the other one and narrows slowly as the sample size grows.
 
 The other safe option is the one from {ref}`numba_ex3`: draw the points before the loop so that the parallel loop only reads from memory.
 
--- a/lectures/numba.md
+++ b/lectures/numba.md
@@ -436,6 +436,17 @@ For GPU-based parallelization, see our {doc}`lectures on JAX <jax_intro>`.
 
 ## Exercises
 
+{ref}`speed_ex1` and {ref}`numba_ex3` both estimate $\pi$ by Monte Carlo from random samples in the unit square.
+
+We generate them here and store them in `u_draws` and `v_draws` so that we can use them in both exercises and compare results
+
+```{code-cell} ipython3
+n = 1_000_000
+rng = np.random.default_rng()
+u_draws = rng.uniform(size=n)
+v_draws = rng.uniform(size=n)
+```
+
 ```{exercise}
 :label: speed_ex1
 
@@ -454,11 +465,6 @@ Compare speed with and without Numba when the sample size is large.
 Here is one solution:
 
 ```{code-cell} ipython3
-n = 1_000_000
-rng = np.random.default_rng()
-u_draws = rng.uniform(size=n)
-v_draws = rng.uniform(size=n)
-
 @jit
 def calculate_pi(u_draws, v_draws):
     n = len(u_draws)
@@ -686,11 +692,6 @@ For the size of the Monte Carlo simulation, use something substantial, such as
 Here is one solution:
 
 ```{code-cell} ipython3
-n = 1_000_000
-rng = np.random.default_rng()
-u_draws = rng.uniform(size=n)
-v_draws = rng.uniform(size=n)
-
 @jit(parallel=True)
 def calculate_pi(u_draws, v_draws):
     n = len(u_draws)
--- a/lectures/numba.md
+++ b/lectures/numba.md
@@ -490,6 +490,48 @@ If we switch off JIT compilation by removing `@jit`, the code takes around
 
 So we get a speed gain of 2 orders of magnitude by adding four characters.
 
+The solution above takes one of two natural approaches: it *draws all the
+random points first*, stores them in `u_draws` and `v_draws`, and then lets the
+jitted function loop over them.
+
+The other approach is to *draw each point inside the loop*.
+
+To do this with a NumPy `Generator`, we pass `rng` in as an argument and call `rng.uniform()` inside the loop body
+
+```{code-cell} ipython3
+@jit
+def calculate_pi_in_loop(rng, n):
+    count = 0
+    for i in range(n):
+        u, v = rng.uniform(), rng.uniform()
+        d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
+        if d < 0.5:
+            count += 1
+    return (count / n) * 4
+```
+
+```{code-cell} ipython3
+with qe.Timer():
+    calculate_pi_in_loop(rng, n)
+```
+
+In this serial setting the two approaches give equally good estimates and run at
+similar speed, but they are not equivalent in *memory use*. 
+
+The first approach
+must hold all $2n$ draws in memory at once --- two arrays of `n` floating point
+numbers, or about `16n` bytes (around $1.6$ GB when `n = 100_000_000`). 
+
+The
+second draws each point on demand and discards it, so its memory footprint does
+not grow with `n`.
+
+This might suggest that drawing inside the loop is the better default.
+
+ But as we
+will see in {ref}`numba_ex_race`, drawing inside the loop interacts
+badly with parallelization.
+
 ```{solution-end}
 ```
 
@@ -685,6 +727,187 @@ a factor of 2 or 3.
 (If you are executing locally, you will get different numbers, depending mainly
 on the number of CPUs on your machine.)
 
+Notice that we drew all of the random points *before* the loop and passed them in
+as arrays, so the parallel loop only *reads* from memory.
+
+Drawing the points *inside* the parallel loop instead is surprisingly delicate.
+
+
+We investigate why, and how to do it safely, in
+{ref}`numba_ex_race`.
+
+```{solution-end}
+```
+
+
+```{exercise}
+:label: numba_ex_race
+
+In {ref}`numba_ex3` we drew all of the random points *before* the parallel loop.
+
+It is tempting to instead draw each point *inside* the `prange` loop, by passing a generator `rng` in as an argument and calling `rng.uniform()` in the loop body.
+
+Try it: the code should run and return a number close to $\pi$, yet there is a subtle bug in this approach.
+
+Investigate as follows:
+
+1. Call your function a few times with the *same* seed and check whether the result is reproducible.
+2. Repeat the estimate many times across a range of sample sizes and compare its spread against a correct parallel version.
+
+Then explain what goes wrong and give a correct way to draw inside a parallel loop.
+
+Hint: try to use a legacy random function such as `np.random.uniform()` instead of a `Generator` and see what happens.
+```
+
+```{solution-start} numba_ex_race
+:class: dropdown
+```
+
+Here is the tempting version.
+
+We pass `rng` in as an argument and call it inside the `prange` loop.
+
+```{code-cell} ipython3
+n = 1_000_000
+rng = np.random.default_rng()
+
+@jit(parallel=True)
+def calculate_pi_in_loop(rng, n):
+    count = 0
+    for i in prange(n):
+        u, v = rng.uniform(), rng.uniform()
+        d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
+        if d < 0.5:
+            count += 1
+    return (count / n) * 4
+
+calculate_pi_in_loop(rng, n)
+```
+
+The code runs without error and returns something close to $\pi$.
+
+But something is silently wrong with the results.
+
+Here, every thread draws from the *same* generator `rng`.
+
+A generator produces each number by updating an internal state.
+
+Under `prange`, many threads read and update that single state at once, with no coordination between them.
+
+This is a [**data race**](https://docs.oracle.com/cd/E19205-01/820-0619/geojs/index.html).
+
+It creates correlations across the draws and can even cause some draws to be duplicated in an
+unpredictable way.
+
+Two symptoms reveal the problem.
+
+*Symptom 1: the result is no longer reproducible.*
+
+A correct generator returns the same answer whenever it is given the same seed.
+
+Because of the data race, the order in which threads happen to touch the shared state affects the stream of draws, so the answer is not reproducible even when the seed is fixed.
+
+```{code-cell} ipython3
+for seed in (1, 1, 1):
+    print(calculate_pi_in_loop(np.random.default_rng(seed), n))
+```
+
+Each call uses the same seed, yet the answers differ.
+
+*Symptom 2: the estimator is far noisier than it should be.*
+
+The duplicated and correlated draws carry less information than $n$ independent draws, so the *effective* sample size is much smaller than $n$.
+
+The fix is to give each thread its own random state, which NumPy's legacy functions such as `np.random.uniform()` do automatically under Numba.
+
+```{code-cell} ipython3
+@jit(parallel=True)
+def calculate_pi_legacy(n):
+    count = 0
+    for i in prange(n):
+        u, v = np.random.uniform(0, 1), np.random.uniform(0, 1)
+        d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
+        if d < 0.5:
+            count += 1
+    return (count / n) * 4
+```
+
+To see the cost of the race, we repeat each estimate many times and plot its spread against the correct version as the sample size grows.
+
+```{code-cell} ipython3
+sample_sizes = np.logspace(3, 6, 10).astype(int)
+num_reps = 20
+
+methods = [("per-thread state (correct)",
+            lambda n: calculate_pi_legacy(n), 'C0'),
+           ("shared generator in prange (data race)",
+            lambda n: calculate_pi_in_loop(np.random.default_rng(), n), 'C1')]
+
+fig, ax = plt.subplots()
+for label, estimate, color in methods:
+    draws = np.array([[estimate(int(m)) for _ in range(num_reps)]
+                      for m in sample_sizes])
+    means, stds = draws.mean(axis=1), draws.std(axis=1)
+    ax.plot(sample_sizes, means, color=color, marker='o', ms=3, label=label)
+    ax.fill_between(sample_sizes, means - 2 * stds, means + 2 * stds,
+                    color=color, alpha=0.2)
+ax.axhline(np.pi, color='k', lw=0.8, ls='--', label=r'$\pi$')
+ax.set_xscale('log')
+ax.set_xlabel('number of samples')
+ax.set_ylabel(r'estimate of $\pi$')
+ax.legend()
+plt.show()
+```
+
+Both bands are centered on $\pi$, but the band associated with the data race is much wider than the other one and narrows very slowly as the sample size grows.
+
+The other safe option is the one from {ref}`numba_ex3`: draw the points before the loop so that the parallel loop only reads from memory.
+
+```{solution-end}
+```
+
+
+```{exercise}
+:label: numba_ex_draw_speed
+
+We now have two correct ways to estimate $\pi$ in parallel.
+
+One draws all the points *before* the loop, as in {ref}`numba_ex3`.
+
+The other draws them *inside* the loop with legacy functions, as in {ref}`numba_ex_race`.
+
+Compare their speed at `n = 100_000_000`, including the time spent generating the random points.
+```
+
+```{solution-start} numba_ex_draw_speed
+:class: dropdown
+```
+
+We time each approach from start to finish, so the pre-draw version pays for building its arrays.
+
+```{code-cell} ipython3
+n = 100_000_000
+rng = np.random.default_rng()
+
+with qe.Timer():
+    u_draws = rng.uniform(size=n)
+    v_draws = rng.uniform(size=n)
+    calculate_pi(u_draws, v_draws)
+```
+
+```{code-cell} ipython3
+with qe.Timer():
+    calculate_pi_legacy(n)
+```
+
+Drawing inside the loop is much faster.
+
+The pre-draw version generates its two arrays on a single thread before the loop begins.
+
+The in-loop version spreads the random number generation across all threads instead.
+
+It also avoids allocating two arrays of `n` numbers, so it saves both time and memory.
+
 ```{solution-end}
 ```
 
@@ -708,8 +931,8 @@ where
 
 1. $\beta$ is a discount factor,
 2. $n$ is the expiry date,
-2. $K$ is the strike price and
-3. $\{S_t\}$ is the price of the underlying asset at each time $t$.
+3. $K$ is the strike price and
+4. $\{S_t\}$ is the price of the underlying asset at each time $t$.
 
 Suppose that `n, β, K = 20, 0.99, 100`.
 
--- a/lectures/numba.md
+++ b/lectures/numba.md
@@ -438,7 +438,7 @@ For GPU-based parallelization, see our {doc}`lectures on JAX <jax_intro>`.
 
 {ref}`speed_ex1` and {ref}`numba_ex3` both estimate $\pi$ by Monte Carlo from random samples in the unit square.
 
-We generate them here and store them in `u_draws` and `v_draws` so that we can use them in both exercises and compare results
+We generate them here and store them in `u_draws` and `v_draws` so that we can use them in both exercises and compare results.
 
 ```{code-cell} ipython3
 n = 1_000_000
@@ -491,10 +491,10 @@ with qe.Timer():
     calculate_pi(u_draws, v_draws)
 ```
 
-If we switch off JIT compilation by removing `@jit`, the code takes around
-150 times as long on our machine.
+If we switch off JIT compilation by removing `@jit`, the code takes
+dramatically longer on our machine.
 
-So we get a speed gain of 2 orders of magnitude by adding four characters.
+So we get a large speed gain by adding four characters.
 
 The solution above takes one of two natural approaches: it *draws all the
 random points first*, stores them in `u_draws` and `v_draws`, and then lets the
@@ -521,20 +521,23 @@ with qe.Timer():
     calculate_pi_in_loop(rng, n)
 ```
 
+```{code-cell} ipython3
+with qe.Timer():
+    calculate_pi_in_loop(rng, n)
+```
+
 In this serial setting the two approaches give equally good estimates and run at
 similar speed, but they are not equivalent in *memory use*. 
 
-The first approach
-must hold all $2n$ draws in memory at once --- two arrays of `n` floating point
+The first approach must hold all $2n$ draws in memory at once --- two arrays of `n` floating point
 numbers, or about `16n` bytes (around $1.6$ GB when `n = 100_000_000`). 
 
-The
-second draws each point on demand and discards it, so its memory footprint does
+The second draws each point on demand and discards it, so its memory footprint does
 not grow with `n`.
 
 This might suggest that drawing inside the loop is the better default.
 
- But as we
+But as we
 will see in {ref}`numba_ex_race`, drawing inside the loop interacts
 badly with parallelization.
 
@@ -693,7 +696,7 @@ Here is one solution:
 
 ```{code-cell} ipython3
 @jit(parallel=True)
-def calculate_pi(u_draws, v_draws):
+def calculate_pi_parallel(u_draws, v_draws):
     n = len(u_draws)
     count = 0
     for i in prange(n):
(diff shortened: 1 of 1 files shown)


lectures/numba.md
```diff
@@ -636,11 +640,17 @@ For the size of the Monte Carlo simulation, use something substantial, such as
...
-def calculate_pi(n=1_000_000):
+def calculate_pi(u_draws, v_draws):
+    n = len(u_draws)
     count = 0
     for i in prange(n):
-        u, v = np.random.uniform(0, 1), np.random.uniform(0, 1)
+        u, v = u_draws[i], v_draws[i]
         d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
         if d < 0.5:
             count += 1
```
HumphreyYang: I think this avoids the thread-safety issue, but the timing comparison is no longer quite apples-to-apples because the random draws have been moved outside the timed function.

Personally, I think we can generate the draws in the first solution by moving

```python
n = 1_000_000
rng = np.random.default_rng()
u_draws = rng.uniform(size=n)
v_draws = rng.uniform(size=n)
```

to line 457, and modifying the function in the code block starting at line 457 (solution to exercise 1) to take `u_draws` and `v_draws`, like the function here.

In this case, we would have a fair timing comparison.

Please let me know what you think!

HumphreyYang (commented):
Many thanks @Chihiro2000GitHub! It looks good to me. Just one small comment!

lectures/numba.md
HumphreyYang: Hi @mmcky, I think this is an example where we need to make a tough choice between keeping the legacy API for simplicity and using the new `rng` API with some trade-offs.

mmcky:
Thanks @Chihiro2000GitHub for the careful classification and write-up. As @HumphreyYang says most of the PR looks good. A few things to resolve before merging:

## 1. Generator passed into `@jit` (Case B, `speed_ex1`)

I'd like to confirm that this stays in nopython mode before we merge. Numba's support for `numpy.random.Generator` objects is limited and version-dependent — "runs" isn't the same as "stays in nopython mode," and if it silently falls back, the lecture's claim immediately below the code cell breaks:

> If we switch off JIT compilation by removing `@jit`, the code takes around 150 times as long on our machine.

Could you re-run with `@jit(nopython=True)` (which raises rather than falls back) and report the with-JIT vs. without-JIT timing? If the 150x figure no longer holds, we should change approach.

## 2. Apples-to-apples timing (Case D vs Case B) — agree with @HumphreyYang

Humphrey's point on `numba_ex3` is right: lifting the draws out of the timed function in Case D but leaving them inside in Case B makes the parallel-vs-serial comparison unfair. I'd suggest **applying the "lift draws outside" pattern to `speed_ex1` as well**. 

Two benefits:

- The timing comparison across the two exercises becomes consistent.
- It sidesteps issue 1 entirely — the JITed function only sees plain arrays, no `Generator` object crossing the JIT boundary.

This also gives readers a clean, consistent rule: **JIT-compiled hot loops consume pre-drawn arrays.**

## 3. Same concern in Case C (`speed_ex2`)

`compute_series_numba = jit(compute_series)` JIT-compiles a function that calls `rng.uniform(...)` internally. Same nopython-mode question as issue 1. Since `rng.uniform(0, 1, size=n)` is called once to produce the array `U`, the cleanest fix is symmetric with issue 2: draw `U` outside and pass it in as an argument. That keeps the JITed body free of `Generator` objects.

## 4. Minor — comment placement in Case E

The comment inside the inner loop explaining why draws are kept inline would read more naturally as a one-liner above the `for t in range(n):` loop, or in the prose just before the code cell. Not blocking.

---

**Summary of requested changes:** confirm nopython mode for Cases B and C (or refactor them to take pre-drawn arrays, which I'd prefer), and adopt the same pre-drawn-array pattern for `speed_ex1` so the timing comparison with `numba_ex3` is fair. Cases A and E are good as-is. 

Thanks again!

Also this is definitely the hardest one I think :-) Thanks for iterating on this.

lectures/numba.md
mmcky: Thanks @HumphreyYang I agree with you. I have added some additional notes below regarding double checking `nopython` mode as well.

Chihiro2000GitHub:
Thanks for the detailed review, @mmcky and @HumphreyYang! Here is a summary of the changes made and a timing verification.

## Changes to `numba.md`

- **Case B (`speed_ex1`) — Issues 1 & 2**: Pre-draw `u_draws` and `v_draws` outside the `@jit` function and pass them as arguments, matching the pattern already used in `numba_ex3`. The `rng.Generator` object no longer crosses the JIT boundary, and the timing comparison with `numba_ex3` is now apples-to-apples.
- **Case C (`speed_ex2`) — Issue 3**: Draw `U` outside `compute_series` and pass it as an argument, keeping the JIT-compiled body free of `Generator` objects.
- **Case E (`numba_ex4`) — Issue 4**: Moved the inline comment about pre-allocation trade-offs to the prose just before the code cell.

## Timing verification (`@jit(nopython=True)`)

I chose the refactor approach, but also ran `@jit(nopython=True)` as requested in Issue 1 to confirm nopython mode and check the timing:

<details>
<summary>Click to expand</summary>

```python
"""
Temporary test: verify refactored Cases B and C in numba.md.
- Case B: @jit(nopython=True) timing, old API vs new API
- Case C: @jit(nopython=True) timing, new API only
"""
import numpy as np
import quantecon as qe
from numba import jit

# ---------------------------------------------------------------------------
# Case B: speed_ex1  (pi approximation)
# ---------------------------------------------------------------------------
print("=== Case B: calculate_pi ===")

# --- Old API ---
@jit(nopython=True)
def calculate_pi_old_jit(n=1_000_000):
    count = 0
    for i in range(n):
        u, v = np.random.uniform(0, 1), np.random.uniform(0, 1)
        d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
        if d < 0.5:
            count += 1
    area_estimate = count / n
    return area_estimate * 4

def calculate_pi_old_pure(n=1_000_000):
    count = 0
    for i in range(n):
        u, v = np.random.uniform(0, 1), np.random.uniform(0, 1)
        d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
        if d < 0.5:
            count += 1
    area_estimate = count / n
    return area_estimate * 4

print("  [Old API] Warm-up:")
with qe.Timer():
    calculate_pi_old_jit()

print("  [Old API] Timed @jit(nopython=True):")
with qe.Timer() as t_old_jit:
    calculate_pi_old_jit()

print("  [Old API] Timed pure Python:")
with qe.Timer() as t_old_pure:
    calculate_pi_old_pure()

print(f"  Old API speedup: {t_old_pure.elapsed / t_old_jit.elapsed:.1f}x")

# --- New API ---
n = 1_000_000
rng = np.random.default_rng()
u_draws = rng.uniform(size=n)
v_draws = rng.uniform(size=n)

@jit(nopython=True)
def calculate_pi_new_jit(u_draws, v_draws):
    n = len(u_draws)
    count = 0
    for i in range(n):
        u, v = u_draws[i], v_draws[i]
        d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
        if d < 0.5:
            count += 1
    area_estimate = count / n
    return area_estimate * 4

def calculate_pi_new_pure(u_draws, v_draws):
    n = len(u_draws)
    count = 0
    for i in range(n):
        u, v = u_draws[i], v_draws[i]
        d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
        if d < 0.5:
            count += 1
    area_estimate = count / n
    return area_estimate * 4

print("  [New API] Warm-up:")
with qe.Timer():
    calculate_pi_new_jit(u_draws, v_draws)

print("  [New API] Timed @jit(nopython=True):")
with qe.Timer() as t_new_jit:
    calculate_pi_new_jit(u_draws, v_draws)

print("  [New API] Timed pure Python:")
with qe.Timer() as t_new_pure:
    calculate_pi_new_pure(u_draws, v_draws)

print(f"  New API speedup: {t_new_pure.elapsed / t_new_jit.elapsed:.1f}x")

# ---------------------------------------------------------------------------
# Case C: speed_ex2  (Markov chain simulation)
# ---------------------------------------------------------------------------
print("\n=== Case C: compute_series ===")

p, q = 0.1, 0.2
n = 1_000_000
rng = np.random.default_rng()
U = rng.uniform(0, 1, size=n)

def compute_series(n, U):
    x = np.empty(n, dtype=np.int64)
    x[0] = 1
    for t in range(1, n):
        current_x = x[t-1]
        if current_x == 0:
            x[t] = U[t] < p
        else:
            x[t] = U[t] > q
    return x

x = compute_series(n, U)
print(np.mean(x == 0))

with qe.Timer() as t_c_pure:
    compute_series(n, U)

print(f"  Pure Python time: {t_c_pure.elapsed:.4f}s")

compute_series_numba = jit(compute_series, nopython=True)

x = compute_series_numba(n, U)
print(np.mean(x == 0))

with qe.Timer() as t_c_jit:
    compute_series_numba(n, U)

print(f"  JIT time:         {t_c_jit.elapsed:.4f}s")
```

</details>

**Results on my machine (`speed_ex1` / Case B):**

| | Speedup |
|---|---|
| Old API (`np.random.uniform` inside loop), `@jit(nopython=True)` vs pure | ~150x |
| New API (pre-drawn arrays), `@jit(nopython=True)` vs pure | ~400–700x (variable across runs) |

Both Cases B and C compiled successfully under `@jit(nopython=True)` — no object-mode fallback.

## Question about "~150 times"

The old API gives a speedup consistent with the existing "around 150 times" claim. However, after the refactor (pre-drawn arrays), the speedup is considerably larger, so the current text no longer holds. **Would you like me to update that line?** Options I see:

1. Replace with "significantly faster"
2. Replace with "hundreds of times faster"
3. Something else you prefer

You can also reproduce these results by running the code above.

## One more question: `numba_ex3` timing claim

The solution to `numba_ex3` states: *"On our workstation, we find that parallelization increases execution speed by a factor of 2 or 3."* Since this PR also modified the `numba_ex3` solution to use pre-drawn arrays, should we re-verify this figure under the new code as well?

lectures/numba.md
```diff
@@ -552,12 +556,13 @@ p, q = 0.1, 0.2  # Prob of leaving low and high state respectively
...
+U = rng.uniform(0, 1, size=n)
 
-def compute_series(n, rng):
+def compute_series(n, U):
     x = np.empty(n, dtype=np.int64)
     x[0] = 1  # Start in state 1
-    U = rng.uniform(0, 1, size=n)
     for t in range(1, n):
         current_x = x[t-1]
         if current_x == 0:
```
HumphreyYang, line 633: Many thanks @Chihiro2000GitHub,

I think we do not need to move the sampling outside here. 

We can use RNG inside `@jit`-decorated functions as long as we are not using `@jit(parallel=True)`.

(See https://manual.quantecon.org/styleguide/code.html#numba-njit-jit)

So the [old commit](https://github.com/QuantEcon/lecture-python-programming/pull/550/changes/397a8919a225e887e44aa8e2a0517a651e40a1e6), which passes `rng` as an argument, looks good to me for this case which we did not use `@jit(parallel=True)`.

We only need to make Exercises 1 and 3 consistent in how they handle sampling so that the timing comparisons are fair.

lectures/numba.md
```diff
@@ -760,6 +764,9 @@ $$
 
 Using this fact, the solution can be written as follows.
 
+Note that random draws are kept inside the inner loop rather than pre-allocated,
+to avoid creating large shock arrays of size `M * n`.
+
```
HumphreyYang: This is a nice clarification! It would be even better to explicitly state that we are using the legacy API.

````
```{note}
Note that random draws are kept inside the inner loop and do not use `rng` as before.

Instead, we use the legacy `np.random.randn()` API.

This is because Numba support for `Generator` objects is not [thread-safe](https://numba.readthedocs.io/en/stable/reference/numpysupported.html#generator-objects) under parallel execution.
```
````

File: lectures/numba.md
Comment (HumphreyYang) [inline lectures/numba.md]: Here we can reuse  

```
n = 1_000_000
rng = np.random.default_rng()
u_draws = rng.uniform(size=n)
v_draws = rng.uniform(size=n)
```

defined in the exercise 1.

**Review by HumphreyYang — COMMENTED** [review]
Many thanks @Chihiro2000GitHub,

Please find my minor comments attached!

**Comment by Chihiro2000GitHub** [comment]
Thanks so much for taking another look, @HumphreyYang!

This is very helpful.

Since Matt has also been involved in this discussion, I'll wait to see if he has any further thoughts before making the next update, so that I can address both comments consistently,

@mmcky, I'd be grateful for your thoughts when you have time — no rush at all.

**Comment by mmcky** [comment]
Thanks @Chihiro2000GitHub and @HumphreyYang — and agreed this is the tricky one. Here's where I land so we can converge:

**Non-parallel cases (B `speed_ex1`, C `speed_ex2`):** let's keep the pre-drawn-array approach as you have it now, rather than moving RNG back inside the `@jit` functions. The rule "JIT-compiled hot loops consume pre-drawn arrays" keeps every JIT body free of `Generator` objects, guarantees nopython, and sidesteps Numba's version-dependent `Generator`-in-nopython support entirely. @HumphreyYang — I know you noted these are safe without `parallel=True`; I'd still prefer the uniform rule for robustness and so B and D stay apples-to-apples. Happy to discuss if you feel strongly.

**`numba_ex3` (D):** let's keep its draws self-contained (redefined in the block) rather than reusing ex1's, to match our "self-contained solution blocks" convention and avoid a cross-dropdown `NameError` if a reader opens only this solution. Timing fairness needs the same *size*, not the same *values*.

**`numba_ex4` (E):** please upgrade the prose note to a `{note}` admonition that states both reasons — thread-safety under `parallel=True` *and* memory:

````
```{note}
Here we keep the random draws inside the inner loop and use the legacy
`np.random.randn()` API rather than a `Generator`.

This is because Numba's support for `Generator` objects is not
[thread-safe](https://numba.readthedocs.io/en/stable/reference/numpysupported.html#generator-objects)
under parallel execution (`@jit(parallel=True)`).

Pre-drawing the shocks into arrays of shape `(M, n)` would avoid this but is
impractical here, since `M = 10_000_000` would require several GB of memory.
```
````

**Timing claims:** yes, please update the "around 150 times" / "2 orders of magnitude" line under `speed_ex1` — with pre-drawn arrays the speedup is much larger. Could you re-run on your machine and use the measured figure (e.g. "more than two orders of magnitude")? And while you're at it, please re-verify the "factor of 2 or 3" parallel claim in `numba_ex3` under the new code.

Everything else looks good. Thanks for the careful iteration on this one!

Since there are a couple of judgment calls here where @HumphreyYang and I lean slightly differently, it might be quickest to sort these out on a short Zoom — @HumphreyYang, could we grab 15 minutes to align, then summarise the outcome back here for @Chihiro2000GitHub?

**Comment by HumphreyYang** [comment]
Hi @mmcky, I am free to meet anytime now. Please feel free to share a link to me!

**Comment by HumphreyYang** [comment]
Many thanks @mmcky for the meeting! It was really fun. The code fails very silently and subtly.

Please take a look at the build and let me know what you think!

**Comment by mmcky** [comment]
> Many thanks @mmcky for the meeting! It was really fun. The code fails very silently and subtly.
> 
> Please take a look at the build and let me know what you think!

Thanks @HumphreyYang -- this is really interesting! 
So our prior was right -- it would run but present bad results (that look like real results). 

My main comment is I think we should pull 

```python
n = 1_000_000
rng = np.random.default_rng()
u_draws = rng.uniform(size=n)
v_draws = rng.uniform(size=n)
```

to be in a code block before exercise 1 starts -- to make it clear that it is re-used across the exercises. 

I'll have a think about the best way to simplify the 4th exercise. I think this is a REALLY valuable example.

mmcky:
@Chihiro2000GitHub when you have time please review @HumphreyYang suggestion and let us know if this makes sense to you on a fresh read.

Chihiro2000GitHub:
Hi @mmcky and @HumphreyYang 

Thank you so much for your careful and valuable reviews. I really appreciate them.

I'll go through your suggestions after I finish checking the chapter I'm currently working on in the new book.

Thanks  again😊

mmcky:
@HumphreyYang -- great discussion and nice updates re: spread to demonstrate this problem. It's a really nice example of where you can get a result that looks right but isn't when working on programs. I think a valuable lesson. I think the final question comes down to where we should have this -- here or in a `Programming for Economists` lessons type guide.

---

@jstac — looping you in on PR #550 while it's still in review; would value your comments on the scope before we converge. It began as a mechanical np.random → Generator migration in numba.md (QuantEcon/meta#299) but has grown into a substantial pedagogical review of [our current style guide](https://manual.quantecon.org/styleguide/code.html#numpy-random-number-generation).

What drove the expansion. Migrating to a Generator exposed a subtle bug: sharing one Generator across a Numba prange parallel loop is a silent data race. The code runs and returns something close to π — it looks correct — but it's (1) not reproducible even with a fixed seed, and (2) far noisier, because threads clobber the generator's shared state, giving correlated/duplicated draws and a much smaller effective sample size. Humphrey confirmed this in a build; it fails quietly enough that a reader would never notice.

**Relationship to the style guide:** This is essentially our existing "don't call rng.* inside a prange loop" warning turned into a full, worked lesson — the guide asserts the corruption; the new exercises show it. The migrated patterns (speed_ex1, speed_ex2, numba_ex3) all follow the guide; numba_ex3 is the guide's canonical "draw outside, pass arrays in" pattern. The PR also extends past the guide in two ways, which we'd fold into the style guide at merge:

The guide gives only one parallel-safe option (draw outside). The PR adds a second: legacy np.random.* inside a prange loop is thread-safe under Numba (each thread gets its own state) — so "legacy = avoid" needs a carve-out for parallel loops. This is the recommended route in the memory-constrained numba_ex4.
The guide says draw-outside is "usually faster." The new numba_ex_draw_speed finds the opposite at n=100M — in-loop legacy is much faster (RNG spread across threads, no big array allocation) — so that claim needs qualifying.

**New content (the scope to weigh):** two new linked exercises — numba_ex_race (the race, its two symptoms, the per-thread-state fix, and a spread plot) and numba_ex_draw_speed (pre-draw vs in-loop legacy at n=100M) — plus a discussion block after speed_ex1 on the draw-first/draw-in-loop memory tradeoff.

**Where we'd welcome your input:** this turns a foundational π-Monte-Carlo exercise into a three-exercise treatment of parallel-RNG data races. Does that depth and placement feel right for this lecture, or would you trim or move some of it?

(Heads-up, not blocking: the "~150×" / "2 orders of magnitude" / "factor of 2 or 3" timing claims still need re-measuring under the new code — @Chihiro2000GitHub  to do. )

Chihiro2000GitHub:
@mmcky -- Thanks for the detailed summary. I agree that the new exercises look like an interesting and potentially valuable addition. Personally, I can see the pedagogical value in including them, but the question of whether this scope is appropriate for this lecture seems beyond what I can confidently assess on my own.

For that reason, I also agree that it would be best to hear @jstac’s view on the scope and placement, especially given his role as the author of the lecture.

Once the final version of the code and structure is settled, I’ll do another final check, including re-measuring the “~150×” / “2 orders of magnitude” / “factor of 2 or 3” timing claims under the new code.

Thanks again.

jstac (commented):
Thanks for this — the data-race teaching addition is genuinely valuable, and I verified the core mechanics hold on **numba 0.61.0 / numpy 2.1.3**: a `Generator` passed into a nopython `@jit` works, a shared `Generator` under `prange` compiles and runs, and the same-seed results really do differ (the data race is real and reproducible). 

My comments are mostly non-blocking, but two are worth a deliberate decision before merge:

- **Hard-coded timing numbers** ("~150×", "2 orders of magnitude", "factor of 2 or 3"). The `speed_ex1` rewrite moved the RNG calls out of the loop body, so the no-JIT baseline behind "~150×" almost certainly changed — these should be re-measured and, ideally, replaced with qualitative language so they don't drift on future toolchain bumps.
- **Non-reproducible published output** in `numba_ex_race` — every CI build will render different numbers/figure. Needs a maintainer call.

Inline notes below cover the timing-comparison asymmetry, name-reuse fragility, and a few prose/formatting fixes.

lectures/numba.md
```diff
@@ -436,6 +436,17 @@ For GPU-based parallelization, see our {doc}`lectures on JAX <jax_intro>`.
 
 ## Exercises
 
+{ref}`speed_ex1` and {ref}`numba_ex3` both estimate $\pi$ by Monte Carlo from random samples in the unit square.
+
+We generate them here and store them in `u_draws` and `v_draws` so that we can use them in both exercises and compare results
```
jstac: **Pedagogical placement + nit.** Generating `u_draws`/`v_draws` here, at the top of the Exercises section *before* `speed_ex1` is even posed, partly gives away the approach the student is asked to discover. Consider introducing the shared draws closer to where they're first consumed (or framing them as "we'll reuse these in the solutions").

Also a missing period at the end of this line: "...compare results**.**"

lectures/numba.md
```diff
@@ -471,19 +483,61 @@ Now let's see how fast it runs:
...
+    calculate_pi(u_draws, v_draws)
 ```
 
 ```{code-cell} ipython3
 with qe.Timer():
-    calculate_pi()
+    calculate_pi(u_draws, v_draws)
 ```
 
 If we switch off JIT compilation by removing `@jit`, the code takes around
```
jstac: **Hard-coded timing number is likely stale now.** The old `calculate_pi` drew randoms *inside* the loop (`np.random.uniform(...)` calls); the new version just **indexes into pre-drawn arrays**. Removing the RNG calls from the loop body changes the no-JIT Python baseline substantially, so the "~150×" ratio this sentence quotes — and the "2 orders of magnitude" claim two lines down — were measured against the *old* code and are probably no longer accurate.

More broadly, hard-coding exact multipliers into the lecture is fragile: they drift silently on every NumPy/Numba/hardware bump. Recommend re-measuring under the new code **and** softening to qualitative language, e.g. "more than two orders of magnitude faster on our machine" / "dramatically faster", rather than a precise figure.

lectures/numba.md
```diff
@@ -471,19 +483,61 @@ Now let's see how fast it runs:
...
+        u, v = rng.uniform(), rng.uniform()
+        d = np.sqrt((u - 0.5)**2 + (v - 0.5)**2)
+        if d < 0.5:
+            count += 1
+    return (count / n) * 4
+```
+
+```{code-cell} ipython3
+with qe.Timer():
+    calculate_pi_in_loop(rng, n)
```
jstac, line 521: **Unfair timing comparison.** `calculate_pi(u_draws, v_draws)` above is timed **twice** (so the 2nd run excludes JIT compilation), but `calculate_pi_in_loop` is timed **once** — so its single measurement *includes* compilation overhead. The prose then claims the two "run at similar speed," which this methodology can't reliably support. Add an untimed warm-up call first, or time it twice like the other.

lectures/numba.md
```diff
@@ -471,19 +483,61 @@ Now let's see how fast it runs:
...
+random points first*, stores them in `u_draws` and `v_draws`, and then lets the
+jitted function loop over them.
+
+The other approach is to *draw each point inside the loop*.
+
+To do this with a NumPy `Generator`, we pass `rng` in as an argument and call `rng.uniform()` inside the loop body
+
+```{code-cell} ipython3
+@jit
+def calculate_pi_in_loop(rng, n):
```
jstac, line 509: **Name reuse is fragile/confusing.** `calculate_pi_in_loop` is defined here as serial `@jit`, then **redefined** as `@jit(parallel=True)` in `numba_ex_race`. Likewise `calculate_pi` is serial in `speed_ex1` and redefined parallel in `numba_ex3`. A reader jumping between dropdown solutions can't tell which version is live, and `numba_ex_draw_speed` (below) silently depends on the redefinition order. Consider distinct names (`calculate_pi_parallel`, `calculate_pi_in_loop_parallel`, ...).

lectures/numba.md
```diff
@@ -471,19 +483,61 @@ Now let's see how fast it runs:
...
+must hold all $2n$ draws in memory at once --- two arrays of `n` floating point
+numbers, or about `16n` bytes (around $1.6$ GB when `n = 100_000_000`). 
+
+The
+second draws each point on demand and discards it, so its memory footprint does
+not grow with `n`.
+
+This might suggest that drawing inside the loop is the better default.
+
+ But as we
```
jstac: **Formatting.** There's a stray leading space before "But as we" (renders oddly), and the surrounding sentences are split across paragraph boundaries so paragraphs begin with a lone "The" ("The first approach / must hold...", "The / second draws..."). Please reflow these into normal paragraphs without the mid-sentence hard breaks.

lectures/numba.md
```diff
@@ -671,6 +728,187 @@ a factor of 2 or 3.
 (If you are executing locally, you will get different numbers, depending mainly
 on the number of CPUs on your machine.)
 
+Notice that we drew all of the random points *before* the loop and passed them in
```
jstac, line 748: Same hard-coded-number concern as in `speed_ex1`: the "a factor of 2 or 3" speed-up quoted just above (line ~726) is a precise figure baked into the text. It's heavily machine/thread-dependent and easy to leave stale. Suggest qualitative phrasing ("a modest speed-up, depending on your core count").

lectures/numba.md
```diff
@@ -671,6 +728,187 @@ a factor of 2 or 3.
...
+                    color=color, alpha=0.2)
+ax.axhline(np.pi, color='k', lw=0.8, ls='--', label=r'$\pi$')
+ax.set_xscale('log')
+ax.set_xlabel('number of samples')
+ax.set_ylabel(r'estimate of $\pi$')
+ax.legend()
+plt.show()
+```
+
+Both bands are centered on $\pi$, but the band associated with the data race is much wider than the other one and narrows very slowly as the sample size grows.
```
jstac: Two things on this data-race figure:

1. **Non-reproducible published output.** This exercise *intentionally* produces different numbers/figure on every execution. QuantEcon runs notebooks in CI and publishes the rendered HTML, so this cell's output and the plot will change on every build. Worth an explicit maintainer decision (@mmcky / @HumphreyYang) — accept as a deliberate exception, or keep the race in prose while stabilizing the figure.
2. **Wording may overstate the effect.** I reproduced the race on numba 0.61 / numpy 2.1 (8 threads): same-seed results do differ ✅, but the race std was only ~**1.7×** the legacy std at n=1e5, and the legacy std was already near the theoretical ideal. "much wider ... narrows very slowly" may be stronger than the rendered figure shows; the magnitude is thread-count-dependent. Suggest checking the figure on the build hardware and softening if needed.

lectures/numba.md
```diff
@@ -671,6 +728,187 @@ a factor of 2 or 3.
...
+We time each approach from start to finish, so the pre-draw version pays for building its arrays.
+
+```{code-cell} ipython3
+n = 100_000_000
+rng = np.random.default_rng()
+
+with qe.Timer():
+    u_draws = rng.uniform(size=n)
+    v_draws = rng.uniform(size=n)
+    calculate_pi(u_draws, v_draws)
```
jstac: This solution calls `calculate_pi(u_draws, v_draws)` expecting the **parallel** version from `numba_ex3` — but that only holds because `calculate_pi` was redefined further up (see the naming note in `speed_ex1`). The comparison is correct only by top-to-bottom execution order. Renaming the parallel function would make this robust and self-documenting.

jstac:
@mmcky @HumphreyYang @Chihiro2000GitHub 

Sorry for the long message -- obviously Claude generated -- probably best handled in another Claude pass that i'll leave to you guys.

Great work clarifying these issues!

I like the new exercises.  We're moving towards JAX and away from Numba for various reasons --- and one is the careful and intuitive treatment of random numbers in JAX under parallelization.  However, while we're still using Numba, these points are important to clarify.  Much appreciated.

I mentioned above that we should try to avoid specific numbers baked into text, certainly 150x and even phrases like "2 orders of magnitude".  I used to do this but I'm trying to avoid it for obvious reasons.  Let's try to use phrases like "large" or "substantial" and let the code outputs speak for themselves.

(Worth adding to the style guide?)

I'll leave this for you guys to wrap up.

Chihiro2000GitHub:
Thank you @jstac for reviewing this! 

@mmcky @HumphreyYang -- I’ll go through the whole discussion here again and also check whether the timing claims such as “~150×,” “2 orders of magnitude,” and “a factor of 2 or 3” are still valid.

I’m a bit busy at the moment, so I’ll probably tackle this on Thursday or Friday.

I’d appreciate it if we could discuss it further after that!

Thanks again!

jstac:
Thanks @Chihiro2000GitHub 

> I’ll go through the whole discussion here again and also check whether the timing claims such as “~150×,” “2 orders of magnitude,” and “a factor of 2 or 3” are still valid.

Actually, what I'm suggesting is that we avoid quantitative terms like this, that can become incorrect as hardware and software change.  better to just say "much faster" etc.

Chihiro2000GitHub:
Thanks @jstac 

Ah, I see — I misunderstood your comment. Thank you for clarifying, and sorry about that!

jstac:
Thanks @Chihiro2000GitHub , all good.

Chihiro2000GitHub:
Thanks @mmcky, @HumphreyYang and @jstac for all the detailed review — this was a hard one, and I really appreciate it.

I pushed a set of changes for the clear review points.

**Changes in this update**
- Renamed the parallel functions so that the serial version and the parallel version no longer have the same name (`calculate_pi_parallel`, `calculate_pi_in_loop_parallel`). This removes the problem @jstac pointed out, where the version you get depends on the order the cells are run.
- Changed the exact timing numbers to words, as @jstac suggested: the "~150×" / "2 orders of magnitude" in `speed_ex1` and the "factor of 2 or 3" in `numba_ex3` are now written in a general way, so they will not become wrong when hardware or libraries change.
- Made the `speed_ex1` timing fair: `calculate_pi_in_loop` is now timed twice (like `calculate_pi`), so the second run does not include the JIT compilation time.
- Changed the `numba_ex4` inline note into a `{note}` box that gives both reasons (a `Generator` is not thread-safe under `parallel=True`, and pre-drawing the `(M, n)` shocks would use too much memory).
- Fixed the small formatting issues in the memory paragraph (an extra space, and sentences broken in the middle), and added the missing period after "…compare results".

**Two points where I would like your opinion before I continue:**

**1. The data-race figure in `numba_ex_race` changes every time in CI.**
As @jstac said, this cell gives a different result and figure every time it runs, and QuantEcon publishes notebooks that are run in CI. I feel this is a decision for the maintainers, not something I should decide alone. So I would be grateful if @mmcky and @HumphreyYang could first decide whether we keep it as an intended exception, or explain the race in the text but make the figure stable. I will do whatever you decide.

@jstac also asked whether the words "much wider … narrows very slowly" are too strong for what the figure shows. To check this, I need to look at the figure. But which figure we will have depends on the decision above — that is, whether we keep the figure that changes on every run, or make it stable. So I do not think it makes sense to check the wording until that decision is made. For this reason, I will wait on this point too.

**2. Where to put the shared `u_draws` / `v_draws` block (before `speed_ex1`).**
@jstac's point is that defining these before `speed_ex1` is written partly shows the reader the intended approach in advance. I thought about the "we'll reuse these in the solutions" wording, but honestly I do not think it solves this problem — the block is still shown above the exercise, so the reader still sees the approach in advance. That wording mainly makes the reuse clear (which is what @mmcky asked for), but it does not hide the approach.

My personal opinion: showing the approach in advance does bother me, but I think it is better than the confusion we would get by moving the block into a solution dropdown. So I would prefer to keep the block before the Exercises section, as it is now. But showing the approach in advance is still a real problem, so if anyone has a better idea, I am very happy to use it. I could not think of a good one myself.

Thanks again!

jstac:
Thanks @Chihiro2000GitHub — here are the decisions on your two open questions so you can do the final pass.

**1. Non-reproducible output in `numba_ex_race`.** Keep it as is. We only diff the source `.md`, not rendered outputs, so there's no repo-hygiene problem, and the varying result is the point of the exercise. No change needed. We'll accept the small chance of a broken-looking plot in a build and fix it reactively if it ever happens.

**2. Placement of the shared `u_draws`/`v_draws` block.** Keep it where it is, before `speed_ex1`. The existing sentence ("We generate them here and store them in `u_draws` and `v_draws` so that we can use them in both exercises and compare results") already frames it as shared setup, so nothing more is needed.

One small wording change while you're in there: in the data-race spread paragraph, drop "much" and "very" so the text doesn't lean on a figure that varies between runs:

> Both bands are centered on $\pi$, but the band associated with the data race is wider than the other one and narrows slowly as the sample size grows.

Everything else from the earlier rounds looks good. Once that edit is in, this is ready to merge. Thanks again for the careful work on this one.

jstac:
Thanks @Chihiro2000GitHub .   The comment above is written by Claude after we discussed the PR together.  It only asks you to delete two words.

It's best for us to avoid such words anyway --- we should keep our language plain and simple, so each lecture sounds academic.

Many thanks for your hard work on this.  (Thanks also @mmcky and @HumphreyYang for reviews.) 

Once you've made those small edits we'll merge.

Chihiro2000GitHub:
Thanks @jstac for the clarification and review — I've dropped "much" and "very" in the data-race paragraph as suggested.

@mmcky @HumphreyYang @jstac — this was a challenging one, but we finally got it over the line. Thank you all so much for the careful reviews! 😊

mmcky:
Thanks @Chihiro2000GitHub — this has turned into a really valuable lecture and I'm excited to get it merged! 🎉

I re-ran everything locally (numba 0.62.1 / numpy 2.3.5) and it all checks out beautifully — the data race symptoms reproduce exactly as described, and the `numba_ex_draw_speed` result holds strongly (in-loop legacy ~18x faster at n=100M on my machine). Really nice work.

I did spot one small timing asymmetry in the new `speed_ex1` discussion that's worth a final tweak. The prose says the pre-draw and in-loop approaches "run at similar speed", but the timed cells above it show roughly a 20x gap (~0.0004s vs ~0.008s on my machine). The reason: the pre-draw cell times *only the loop* — the draws happen in the shared setup block at the top of the Exercises section and are never timed — while the in-loop cell pays for its draws inside the timed function. So the outputs the reader sees don't quite back up the sentence.

The good news is the prose is *right* — when array generation is included, the two approaches really are comparable (~0.0098s vs ~0.0084s when I measure it). So the fix is just to make the comparison apples-to-apples, the same way `numba_ex_draw_speed` already does — add one end-to-end timed run of the pre-draw approach just before the "similar speed" paragraph:

```python
with qe.Timer():
    u2 = rng.uniform(size=n)
    v2 = rng.uniform(size=n)
    calculate_pi(u2, v2)
```

with a short sentence noting that this run includes the cost of generating the arrays, which the earlier cells excluded. The shared `u_draws`/`v_draws` block stays right where it is — this just adds one fairly-timed run so the comparison the prose makes is the one the reader sees.

Would you mind adding this as the final change when you get a chance? No rush at all — and thanks again for all the careful iteration on this one, it's been a great collaboration. 😊

Chihiro2000GitHub:
One more small thing, @jstac — I realized that in one of my earlier edits I'd written "dramatically longer" in the JIT section. Re-reading it with your point about keeping the language plain and academic in mind, I've now toned it down to "considerably longer" (same spirit as dropping "much"/"very"). Happy to revert if you'd prefer the original.

Chihiro2000GitHub:
Thanks @mmcky for the careful re-run and for catching this! I've added the end-to-end timed run as you suggested — could you take a look when you get a chance (`1359bb4`)? It's been a real pleasure working on this with you all. 😊

jstac:
Great work @Chihiro2000GitHub , many thanks.  Merging!!

mmcky:
### ✅ Translation sync completed (fr)

**Target repo**: QuantEcon/lecture-python-programming.fr
**Translation PR**: https://github.com/QuantEcon/lecture-python-programming.fr/pull/6
**Files synced** (1):
- `lectures/numba.md`

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python-programming.zh-cn/pull/70
**Files synced** (1):
- `lectures/numba.md`

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python-programming.fa/pull/133
**Files synced** (1):
- `lectures/numba.md`

PR #549, opened 2026-05-09
[numpy.md] Update np.random → Generator API

Chihiro2000GitHub:
## Summary

This PR migrates legacy NumPy random API calls in `numpy.md` as part of [QuantEcon/meta#299](https://github.com/QuantEcon/meta/issues/299).
All `np.random.randn`, `np.random.binomial`, and `from numpy.random import uniform` usages are replaced with the Generator API via `np.random.default_rng()`.

## Details

- `rng = np.random.default_rng()` is introduced at the first point of use (Mutability section) and reused throughout the main lecture text.
- Exercise solution blocks (`np_ex3`, `np_ex4`) define their own local `rng` to remain self-contained.
- `np_ex4` (Part 1 and Part 2, exercise and solution) retains seed 123 via `np.random.default_rng(123)`, matching the original intent of reproducible comparison between broadcasting and for-loop results.
- `DiscreteRV` in the `np_ex2` solution stores `self.rng = np.random.default_rng()` in `__init__` and uses `self.rng.uniform(...)` in `draw`.
- No fixed seed was introduced beyond what was already present in the original.
- No Numba-related code was found in this file.
- Local build completed successfully (`jb build lectures`); `numpy.md` executed without errors in 30 seconds.

## Notes for reviewers

One item was intentionally left unchanged and would benefit from reviewer judgment:

The prose in the Sub-packages section currently reads:

> We've already seen how we can generate random variables using `np.random`

After this migration, the code directly below uses `rng.standard_normal()` and `rng.binomial()` rather than `np.random.xxx()` directly. Technically `rng` is still obtained from `np.random.default_rng()`, so the sentence is not wrong — but it may feel slightly mismatched to a reader. Happy to update the wording if you think that would be clearer.

Hi @mmcky and @HumphreyYang, I'd be grateful if you could take a look when you have time.

PR #538, opened 2026-04-20
[python_by_example.md] Update np.random → Generator API

Chihiro2000GitHub:
This PR updates legacy NumPy random API usage in `python_by_example.md` to the recommended Generator API, as part of https://github.com/QuantEcon/meta/issues/299

I introduced a local `rng = np.random.default_rng()` and replaced the relevant legacy random calls using the new Generator API.

In the main lecture flow, I introduced `rng` only once and then reused it in subsequent cells, in order to keep the structure and presentation as close to the original as possible.

For the exercise solutions, I took a slightly different approach and introduced `rng` within each solution block. My reasoning was that these solutions seemed more natural as relatively self-contained answer units, especially since the exercise statements explicitly specify the relevant imports. That said, if this is not the preferred approach, it should be easy to revise by removing those local `rng` definitions.

I also updated the first sentence in Section 3.3.1.3 (“Subpackages”), since it explicitly referred to the legacy line `np.random.randn(100)` and would otherwise have become inconsistent with the revised code.

I built the lecture locally to check that the changes work as expected.

I’d be very glad to hear any thoughts, especially on the choice to define `rng` separately within each exercise solution block.

PR #595, opened 2026-08-03
FIX: lecture cleanup batch from the 2026-08 tracker audit

### LINKED ISSUE #529: Bug: PRNG key reuse in autodiff lecture + unused import in numpy_vs_numba_vs_jax
## Summary

Two minor issues found during translation review of the zh-cn repo (via Copilot code review on [QuantEcon/lecture-python-programming.zh-cn#19](https://github.com/QuantEcon/lecture-python-programming.zh-cn/pull/19)):

### 1. PRNG key reuse in `lectures/autodiff.md`

In the gradient descent / simulated data section, the same `key` is used for both `jax.random.uniform` and `jax.random.split`:

```python
x = jax.random.uniform(key, (n,))

α, β, σ = 0.5, 1.0, 0.1
key, subkey = jax.random.split(key)
ϵ = jax.random.normal(subkey, (n,))
```

This violates JAX's guidance that keys should be treated as single-use. `key` is consumed by `uniform` and then reused as input to `split`, which can lead to correlated draws.

**Suggested fix:**
```python
key, x_key, eps_key = jax.random.split(key, 3)
x = jax.random.uniform(x_key, (n,))

α, β, σ = 0.5, 1.0, 0.1
ϵ = jax.random.normal(eps_key, (n,))
```

### 2. Unused `random` import in `lectures/numpy_vs_numba_vs_jax.md`

```python
import random
```

`random` is imported but never used in the lecture. Can be safely removed.

---

Found by GitHub Copilot code review during automated translation sync.

### LINKED ISSUE #552: Follow-ups from np.random → Generator API migration (rng scoping & docs)
A few code-quality follow-ups noticed across the recent `np.random` → `numpy.random.Generator` migration PRs (#538, #541, #544, #545, #547, #548, #549) — surfaced by Copilot during the corresponding translation-sync reviews in the Farsi target repo. Each is a small fix; grouping in one issue for convenience.

## 1. `python_by_example.md` — stale "three lines" wording

In the *Random Draws* section, the narrative says:

> Returning to our program that plots white noise, the remaining **three** lines after the import statements are

But the snippet now has **four** lines (`rng = np.random.default_rng()` was added). Either update the count or include `rng = ...` in the snippet so the description matches.

## 2. `functions.md` — `rng` referenced in cells where it isn't defined

In *Applications → Random sampling*, several code cells call `rng.standard_normal()` / `rng.uniform()` but don't define `rng` in the same cell. If the reader runs the cell in isolation (or restarts the kernel), they get `NameError: rng is not defined`.

Suggested fix: either add `rng = np.random.default_rng()` to each affected cell, move the RNG setup to the shared imports cell at the top, or pass an RNG into `generate_data()` to avoid hidden cell-order dependencies.

## 3. `scipy.md` — same `rng`-scoping issue

The `linregress` example uses `x = rng.standard_normal(200)` / `y = ... rng.standard_normal(200)` without defining `rng` in the same cell/section. Same fix as above (define locally, or document the dependency).

## 4. `numpy.md` — `DiscreteRV.__init__` docstring missing `seed`

`DiscreteRV.__init__` now accepts a `seed` parameter and initializes `self.rng`:

```python
def __init__(self, q, seed=None):
    \"\"\"
    The argument q is a NumPy array, or array like, nonnegative and sums
    to 1
    \"\"\"
```

The docstring only mentions `q`. Worth documenting what `seed` controls (and noting that `seed=None` yields non-reproducible draws).

## 5. (Bonus) `about_py.md` — inconsistent "Pytorch" vs "PyTorch"

The file uses both "PyTorch" and "Pytorch" in different paragraphs (and the line `Pytorch is just one of several Python libraries for deep learning and AI.` near line 142). Standardize on "PyTorch" everywhere.

## 6. `polars.md` — legacy `np.random.*` API *(added 2026-08-04)*

The Polars lecture (#408, merged 2026-07-30) introduced eight call sites using the legacy API — `np.random.seed` / `np.random.randn` / `np.random.choice` / `np.random.rand` (lines 81, 117, 466, 468–472) — the exact pattern the series-wide Generator migration above removed. Flagged as Finding 3 in the #408 review but never filed as an issue; tracked here so a single cleanup PR covers all six items.

---

Happy to send PRs for any/all of these — flagging here first so the team can decide scope/priority.

### LINKED ISSUE #580: [pandas_panel.md] Use direct raw.githubusercontent.com URLs for the data-lectures datasets
#578 pointed the three dataset reads in `pandas_panel.md` at `https://github.com/QuantEcon/data-lectures/raw/main/lectures/{realwage,countries,employ}.csv`, a URL form that 302-redirects to `raw.githubusercontent.com`.

Direct raw URLs (`https://raw.githubusercontent.com/QuantEcon/data-lectures/main/lectures/....csv`) skip the redirect, are friendlier to strict HTTP clients and mirrored environments, and match the form these same lines used before #578 as well as the convention elsewhere in the lecture series.

All three current URLs do resolve (verified 2026-07-18: one redirect, then 200), so this is a minor robustness/consistency improvement rather than a breakage — and possibly a deliberate choice, in which case feel free to close.

Origin: flagged by Copilot review on the sync PRs QuantEcon/lecture-python-programming.zh-cn#72 and QuantEcon/lecture-python-programming.fa#136. The URLs mirror the English source, so any change belongs here and will propagate to all editions through translation sync.

### LINKED ISSUE #589: polars: Exercise 1 solution lacks the drop_nulls() guard that Exercise 2 has
`lectures/polars.md` — Exercise 1's solution computes percentage change with `first()`/`last()` on columns that can contain nulls, without the `drop_nulls()` guard that Exercise 2's solution uses.

**Exercise 2** (lines 762–764) guards:

```python
*[pl.col(idx).drop_nulls().first().alias(f'{idx}_first') ...
*[pl.col(idx).drop_nulls().last().alias(f'{idx}_last') ...
```

**Exercise 1** (line 688) does not:

```python
((pl.col(tick).last() / pl.col(tick).first() - 1) * 100)
```

`read_data_polars` assembles the combined price table through a chain of `join(..., how='full', coalesce=True)` calls. Tickers with different trading calendars — different listing dates, market holidays — leave nulls at the head or tail of individual columns. `pl.col(tick).first()` then returns null, and the percentage change for that ticker becomes null rather than its actual value.

The `sort('Date')` added to `read_data_polars` in #408 fixes row *ordering*, and the comment at lines 671–672 documents why ("Hence the `sort('Date')` before returning, which any later `first()`/`last()` calculation relies on"). It does not address nulls — which is precisely why Exercise 2 needed a separate `drop_nulls()` guard. Exercise 1 has the same exposure and did not get the same fix.

CI does not catch this: nulls propagate into the bar chart silently rather than raising, so the build passes and the affected bar is simply wrong or absent.

Suggested fix — mirror Exercise 2:

```python
((pl.col(tick).drop_nulls().last() / pl.col(tick).drop_nulls().first() - 1) * 100)
```

Found while validating the translation sync of #408 into the French, Farsi and Simplified Chinese editions. GitHub Copilot flagged it independently on all three translation PRs — QuantEcon/lecture-python-programming.fr#20, QuantEcon/lecture-python-programming.fa#144 and QuantEcon/lecture-python-programming.zh-cn#79 — which is what prompted the check here. Those PRs reproduce the source code byte-identically (verified: 52 code cells, non-comment content identical in all three), so correcting it here propagates to every edition on the next sync rather than needing three separate fixes.

### LINKED ISSUE #594: Minor inconsistency in numpy.md: text mentions tic/toc but code uses qe.Timer
In the "Implicit Multithreading" exercise section of `numpy.md` (Exercise 10.4, Part 2), the instructional text says:

> "For this part of the exercise you can use the tic/toc functions from the quantecon library to time the execution."

But the actual code cell right below it uses `qe.Timer(...)` as a context manager, not `tic()`/`toc()`:

```python
with qe.Timer("Broadcasting operation"):
    B = x / y
```

This looks like a leftover from a previous version of the lecture, before the timing code was migrated from `tic`/`toc` to the `Timer` context manager. The sentence introducing the exercise should probably be updated to match the code, e.g.:

> "For this part of the exercise you can use the `qe.Timer` context manager from the quantecon library to time the execution."

I noticed this while doing a French translation review of this lecture and wanted to flag it separately, since it's an issue with the English source rather than the translation itself.

mmcky:
One batch of small, independently verified fixes from the 2026-08 tracker audit. They are grouped into a single PR because every item touches lecture source, so batching means the `.fa`/`.fr`/`.zh-cn` editions pick everything up in one translation sync instead of five.

Closes #589, closes #580, closes #552, closes #529, closes #594.

## Changes

| Issue | File(s) | Change |
| --- | --- | --- |
| #589 | `polars.md` | Exercise 1 solution now guards `first()`/`last()` with `drop_nulls()`, matching Exercise 2 — without it, tickers on different trading calendars produce null `pct_change` values that flow silently into the bar chart |
| #580 | `pandas_panel.md` | The three `github.com/QuantEcon/data-lectures/raw/main/...` URLs now point directly at `raw.githubusercontent.com`, avoiding the 302 redirect |
| #552 (1) | `python_by_example.md` | The white-noise snippet now includes the `rng = np.random.default_rng()` line and the prose counts four lines, not three (also fixes a "genererate" typo) |
| #552 (2) | `functions.md` | The three `generate_data` cells define `rng` locally, so running a cell in isolation no longer raises `NameError` |
| #552 (3) | `scipy.md` | The `linregress` example defines `rng` in its own cell (matching the pattern already used at the Monte Carlo example later in the lecture) |
| #552 (4) | `numpy.md` | `DiscreteRV.__init__` docstring now documents the `seed` parameter and the non-reproducibility of the `seed=None` default |
| #552 (5) | `about_py.md` | "Pytorch" → "PyTorch" (2 sites) |
| #552 (6) | `polars.md`, `pandas.md` | The eight legacy `np.random.*` call sites in `polars.md` migrated to the Generator API; the mirrored Series cell in `pandas.md` (`np.random.randn(4)`, the one site the audit had not flagged) is migrated too so the twin lectures stay consistent |
| #529 | `autodiff.md` | The simulated-data cell splits the PRNG key once into three single-use keys instead of consuming `key` with `uniform` and then re-splitting it (correlated-draws risk); uses the fix suggested in the issue, with `ϵ_key` named to match the lecture's unicode convention |
| #594 | `numpy.md` | Exercise text now says `qe.Timer()` context manager instead of the retired `tic`/`toc` wording, matching the solution code |

## Notes for review

Random streams change in the edited cells: the polars/pandas Series and DataFrame demo cells now draw from a fresh `Generator`, and the polars benchmark block seeds via `np.random.default_rng(42)` instead of `np.random.seed(42)`. Displayed numbers in the published output will differ from the current build, but nothing pedagogical depends on the specific draws. This is the same trade accepted throughout the series-wide Generator migration (#538–#549).

The `rng`-scoping fixes follow the convention already used by the migrated lectures (`numba.md`, `numpy.md` define `rng` in each cell/section that uses it) rather than introducing a shared top-of-lecture definition.

### DIFF
--- a/lectures/numpy.md
+++ b/lectures/numpy.md
@@ -1252,7 +1252,11 @@ class DiscreteRV:
     def __init__(self, q, seed=None):
         """
         The argument q is a NumPy array, or array like, nonnegative and sums
-        to 1
+        to 1.
+
+        The argument seed sets the seed for the underlying random number
+        generator; with the default seed=None, draws are not reproducible
+        across runs.
         """
         self.q = q
         self.Q = cumsum(q)
@@ -1431,7 +1435,7 @@ print(A)
 
 **Part2**: Move on to replicate the result of the following broadcasting operation. Meanwhile, compare the speeds of broadcasting and the `for` loop you implement.
 
-For this part of the exercise you can use the `tic`/`toc` functions from the `quantecon` library to time the execution. 
+For this part of the exercise you can use the `qe.Timer()` context manager from the `quantecon` library to time the execution. 
 
 Let's make sure this library is installed.
 
--- a/lectures/about_py.md
+++ b/lectures/about_py.md
@@ -139,7 +139,7 @@ popularity of a single Python deep learning library
 
 ```{figure} /_static/lecture_specific/about_py/pytorch_vs_matlab.png
 ```
-Pytorch is just one of several Python libraries for deep learning and AI.
+PyTorch is just one of several Python libraries for deep learning and AI.
 
 
 
@@ -349,7 +349,7 @@ We will discuss the details later in the lecture series, where we cover NumPy in
 While NumPy is still the king of array processing in Python, there are now
 important competitors.
 
-Libraries such as [JAX](https://github.com/jax-ml/jax), [Pytorch](https://pytorch.org/), and [CuPy](https://cupy.dev/) also have
+Libraries such as [JAX](https://github.com/jax-ml/jax), [PyTorch](https://pytorch.org/), and [CuPy](https://cupy.dev/) also have
 built in array types and array operations that can be very fast and efficient.
 
 In fact these libraries are better at exploiting parallelization and fast hardware, as
--- a/lectures/autodiff.md
+++ b/lectures/autodiff.md
@@ -350,11 +350,11 @@ Let's generate some simulated data:
 ```{code-cell} ipython3
 n = 100
 key = jax.random.key(1234)
-x = jax.random.uniform(key, (n,))
+key, x_key, ϵ_key = jax.random.split(key, 3)
+x = jax.random.uniform(x_key, (n,))
 
 α, β, σ = 0.5, 1.0, 0.1  # Set the true intercept and slope.
-key, subkey = jax.random.split(key)
-ϵ = jax.random.normal(subkey, (n,))
+ϵ = jax.random.normal(ϵ_key, (n,))
 
 y = α * x + β + σ * ϵ
 ```
--- a/lectures/functions.md
+++ b/lectures/functions.md
@@ -305,6 +305,8 @@ This is accomplished in the next program
 
 (funcloopprog)=
 ```{code-cell} python3
+rng = np.random.default_rng()
+
 def generate_data(n):
     ϵ_values = []
     for i in range(n):
@@ -334,6 +336,8 @@ This is achieved in the next piece of code.
 
 (funcloopprog2)=
 ```{code-cell} python3
+rng = np.random.default_rng()
+
 def generate_data(n, generator_type):
     ϵ_values = []
     for i in range(n):
@@ -366,6 +370,8 @@ To understand this, consider the following version.
 
 (test_program_6)=
 ```{code-cell} python3
+rng = np.random.default_rng()
+
 def generate_data(n, generator_type):
     ϵ_values = []
     for i in range(n):
--- a/lectures/functions.md
+++ b/lectures/functions.md
@@ -305,8 +305,6 @@ This is accomplished in the next program
 
 (funcloopprog)=
 ```{code-cell} python3
-rng = np.random.default_rng()
-
 def generate_data(n):
     ϵ_values = []
     for i in range(n):
@@ -336,8 +334,6 @@ This is achieved in the next piece of code.
 
 (funcloopprog2)=
 ```{code-cell} python3
-rng = np.random.default_rng()
-
 def generate_data(n, generator_type):
     ϵ_values = []
     for i in range(n):
@@ -370,8 +366,6 @@ To understand this, consider the following version.
 
 (test_program_6)=
 ```{code-cell} python3
-rng = np.random.default_rng()
-
 def generate_data(n, generator_type):
     ϵ_values = []
     for i in range(n):
--- a/lectures/pandas.md
+++ b/lectures/pandas.md
@@ -89,7 +89,8 @@ Let's start with Series.
 We begin by creating a series of four random observations
 
 ```{code-cell} ipython3
-s = pd.Series(np.random.randn(4), name='daily returns')
+rng = np.random.default_rng()
+s = pd.Series(rng.standard_normal(4), name='daily returns')
 s
 ```
 
--- a/lectures/pandas_panel.md
+++ b/lectures/pandas_panel.md
@@ -77,7 +77,7 @@ countries and assign it to `realwage`.
 The dataset can be accessed with the following link:
 
 ```{code-cell} ipython3
-url1 = 'https://github.com/QuantEcon/data-lectures/raw/main/lectures/realwage.csv'
+url1 = 'https://raw.githubusercontent.com/QuantEcon/data-lectures/main/lectures/realwage.csv'
 ```
 
 ```{code-cell} ipython3
@@ -197,7 +197,7 @@ function.
 The dataset can be accessed with the following link:
 
 ```{code-cell} ipython3
-url2 = 'https://github.com/QuantEcon/data-lectures/raw/main/lectures/countries.csv'
+url2 = 'https://raw.githubusercontent.com/QuantEcon/data-lectures/main/lectures/countries.csv'
 ```
 
 ```{code-cell} ipython3
@@ -506,7 +506,7 @@ in Europe by age and sex from [Eurostat](https://ec.europa.eu/eurostat/data/data
 The dataset can be accessed with the following link:
 
 ```{code-cell} ipython3
-url3 = 'https://github.com/QuantEcon/data-lectures/raw/main/lectures/employ.csv'
+url3 = 'https://raw.githubusercontent.com/QuantEcon/data-lectures/main/lectures/employ.csv'
 ```
 
 Reading in the CSV file returns a panel dataset in long format. Use `.pivot_table()` to construct
--- a/lectures/polars.md
+++ b/lectures/polars.md
@@ -464,13 +464,13 @@ a grouped weighted average.
 
 ```{code-cell} ipython3
 n = 5_000_000
-rng = np.random.default_rng(42)
+bench_rng = np.random.default_rng(42)
 
-groups = rng.choice(['A', 'B', 'C', 'D'], n)
-values = rng.standard_normal(n)
-weights = rng.random(n)
-extra1 = rng.standard_normal(n)
-extra2 = rng.standard_normal(n)
+groups = bench_rng.choice(['A', 'B', 'C', 'D'], n)
+values = bench_rng.standard_normal(n)
+weights = bench_rng.random(n)
+extra1 = bench_rng.standard_normal(n)
+extra2 = bench_rng.standard_normal(n)
 
 big_pd = pd.DataFrame({
     'group': groups, 'value': values,
--- a/lectures/polars.md
+++ b/lectures/polars.md
@@ -464,13 +464,13 @@ a grouped weighted average.
 
 ```{code-cell} ipython3
 n = 5_000_000
-bench_rng = np.random.default_rng(42)
+rng = np.random.default_rng(42)
 
-groups = bench_rng.choice(['A', 'B', 'C', 'D'], n)
-values = bench_rng.standard_normal(n)
-weights = bench_rng.random(n)
-extra1 = bench_rng.standard_normal(n)
-extra2 = bench_rng.standard_normal(n)
+groups = rng.choice(['A', 'B', 'C', 'D'], n)
+values = rng.standard_normal(n)
+weights = rng.random(n)
+extra1 = rng.standard_normal(n)
+extra2 = rng.standard_normal(n)
 
 big_pd = pd.DataFrame({
     'group': groups, 'value': values,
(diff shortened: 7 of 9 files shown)


mmcky:
@Chihiro2000GitHub would you have time to cast an eye over these changes. Many of them are `rng` updates.

Chihiro2000GitHub:
Thanks @mmcky!

I've had a quick look through the changes.

I'll do a proper review starting this weekend and get back to you with any comments 😊

mmcky:
Reviewed the `lectures/functions.md` change and reverted it in b13f58b — `functions.md` now matches `main` and drops out of this PR.

To answer the question that prompted the look: the `rng` object *was* used in all three cells the change touched — `rng.standard_normal()` in `funcloopprog`, `rng.uniform(0, 1)` / `rng.standard_normal()` in `funcloopprog2`, and `rng.uniform` passed as the callable at the call site in `test_program_6` (that last one is easy to misread in the diff view, since the function body only calls `generator_type()`).

The problem was that the three added `rng = np.random.default_rng()` lines were redundant rather than unused. `rng` is already bound in the first cell of the "Random Draws" section and notebook state carries it forward, so re-creating it changed nothing at execution time — each `default_rng()` is unseeded either way, so there was no reproducibility argument for it either.

The change was also inconsistent about it: the `draw_new` cell in the exercise solutions still calls `rng.uniform()` with no local constructor, relying on the earlier binding. So the "make each cell self-contained" rationale wasn't applied uniformly.

Given the lecture establishes `rng` once at the top of the section, and these particular cells exist to teach function structure rather than RNG setup, dropping the boilerplate is the cleaner outcome. If we do want self-contained cells as a repo-wide convention, that is worth doing deliberately across every cell that touches `rng` — not partially in one lecture.

lectures/numpy.md
```diff
@@ -1252,7 +1252,11 @@ class DiscreteRV:
     def __init__(self, q, seed=None):
         """
         The argument q is a NumPy array, or array like, nonnegative and sums
-        to 1
+        to 1.
+
+        The argument seed sets the seed for the underlying random number
```
mmcky, line 1257: @jstac what is your view on docstrings in lectures. Should they fully document a function interface (like in a library) or be a concise one sentence description?

mmcky:
Followed up on the remaining `rng` usage across the PR (0b077ba). Two tidy-ups:

**`scipy.md`** — dropped the added `rng = np.random.default_rng()` before the `linregress` example. Same situation as `functions.md`: `rng` is already bound earlier in the lecture and carried forward through notebook state, so the line re-created a generator that was already in scope. `scipy.md` now matches `main` and drops out of this PR.

**`polars.md`** — renamed the seeded benchmark generator to `bench_rng`. Replacing `np.random.seed(42)` with `rng = np.random.default_rng(42)` was the right modern equivalent, but it silently rebound the unseeded `rng` from the Series section to a seeded one with different semantics. Nothing downstream depended on it, so this was cosmetic rather than a bug, but the distinct name makes the benchmark's reproducibility intent explicit.

The other RNG changes all look right and are unchanged:

- `pandas.md` and `polars.md` (Series) — genuine legacy `np.random.randn` to Generator API conversions, and the first `rng` use in each file.
- `python_by_example.md` — the added `rng` line is necessary here, not redundant. That cell re-quotes the program from `(ourfirstprog)` and the prose says "the remaining four lines after the import statements", so `rng = np.random.default_rng()` genuinely is one of those lines. The stale "three lines" count and the line-by-line walkthrough were correctly updated alongside it.
- `numpy.md` — the `seed=None` docstring for `DiscreteRV` is accurate.
- `autodiff.md` — splitting `key` once into three named subkeys is a real fix, not just cosmetic. The original drew `x` from `key` and then split that same `key` for the shocks, reusing key material.

Also confirmed there are no legacy `np.random.*` calls left in any of the nine files. The only surviving mention is the illustrative `from numpy.random import rand, randn` inside the historical SciPy-initialization note in `scipy.md`, which is a non-executed block quoting old SciPy internals and should stay as is.

The working rule I applied, if it is useful as a convention: re-create `rng` when a cell is an independent, copy-pasteable example (new section, exercise solution, standalone demo), and rely on the earlier binding when consecutive cells are successive refinements of one program, where the reader is meant to see only what changed.

mmcky:
Correction to my previous comment: the `bench_rng` rename in `polars.md` has been reverted in 9af710c, so the benchmark cell keeps `rng = np.random.default_rng(42)` as originally written.

Reusing the name is fine here. The rename was a readability preference rather than a fix, it had no effect on output, and nothing downstream of the benchmark cell reads `rng`. The seed is unchanged, so the benchmark data remains reproducible.

The `scipy.md` change from that commit stands: the redundant `rng` re-creation before the `linregress` example is still removed, and `scipy.md` still drops out of this PR.

mmcky:
@Chihiro2000GitHub I had a closer look and I actually removed a number of the `rng` calls -- making use of the general `rng` object from the beginning of the lecture. 

@jstac this is good to go -- bar the final call re: `docstring`.

Chihiro2000GitHub:
Thanks for looking into this @mmcky, and thanks for the clear write-up.

**I agree with your rule.** When one cell continues from the one before it, we only need to define `rng` once, in the first cell of the sequence. But a reader may start from the middle of a lecture, for example at a new section or an exercise solution.  If a cell starts a new topic and does not rely on the cells before it, I think it is better to redefine `rng` there.

You said a convention like this is worth applying deliberately, not partially in one lecture. I worked on the same `rng` migration in all three lecture repos: this one (`lecture-python-programming`), `lecture-python-intro`, and `lecture-python.myst`. So the convention would apply to all three. **Shall I open a follow-up issue in each of the three repos after this PR is merged?** I can go through the lectures that use `rng` and list the cells that need a change.

On `polars.md`: `rng` is first created without a seed, and later re-created with seed 42. Reusing the name is fine, but nothing tells the reader why the seed is there — **would a short comment help?**

```python
rng = np.random.default_rng(42)  # fixed seed so the benchmark data is reproducible
```

The rest looks good to me.

Chihiro2000GitHub (approved):
Approving — see my comment above. The `polars.md` suggestion is optional.

mmcky:
thanks @Chihiro2000GitHub for your review and comments. I'll merge this now.

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python-programming.zh-cn/pull/89
**Files synced** (7):
- `lectures/about_py.md`
- `lectures/autodiff.md`
- `lectures/numpy.md`
- `lectures/pandas.md`
- `lectures/pandas_panel.md`
- `lectures/polars.md`
- `lectures/python_by_example.md`

mmcky:
\translate-resync

Regenerating the fr and fa syncs for this batch from current target mains — fr#30 and fa#150 conflicted after their queue-mates merged and their automatic rebases failed (zh-cn merged its twin as zh-cn#89, so its regeneration should be a no-op).

mmcky:

**Translation PR**: https://github.com/QuantEcon/lecture-python-programming.zh-cn/pull/93
**Files synced** (7):
- `lectures/about_py.md`
- `lectures/autodiff.md`
- `lectures/numpy.md`
- `lectures/pandas.md`
- `lectures/pandas_panel.md`
- `lectures/polars.md`
- `lectures/python_by_example.md`

